PR #16, opened 2016-01-04
Add lowerBound

abarth:
This patch adds a lowerBound function to algorithms.dart to complement
binarySearch. It's common to binary search a sorted list for a item that might
or might not be present. If the item is not present, lowerBound lets you learn
where in the collection it would have been.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 1.2.1
+
+* Add lowerBound to binary search for values that might not be present.
+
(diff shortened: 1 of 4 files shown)


abarth:
@lrhn

abarth:
Pull request updated.  Please take another look.  Thanks!

abarth:
Patch updated.  Would you be willing to land it?  Thanks!

lrhn:
Landed

PR #18, opened 2016-01-27
add a forward constructor in PriorityQueue to HeapPriorityQueue

### LINKED ISSUE #17: PriorityQueue should forward a constructor to HeapPriorityQueue
It's nice that `PriorityQueue` allows for different implementations, but (like the core collection APIs) it should have a factory constructor that forwards to the default implementation. Right now, it's confusing that you can't construct a `PriorityQueue` using `new PriorityQueue()`.

a14n:
Fixes #17

### DIFF
--- a/lib/priority_queue.dart
+++ b/lib/priority_queue.dart
@@ -13,6 +13,19 @@ import "dart:collection" show SplayTreeSet;
  */
 abstract class PriorityQueue<E> {
   /**
+   * Creates an empty [PriorityQueue].
+   *
+   * The created [PriorityQueue] is a plain [HeapPriorityQueue].
+   *
+   * The [comparison] is a [Comparator] used to compare the priority of
+   * elements. An element that compares as less than another element has
+   * a higher priority.
+   *
+   * If [comparison] is omitted, it defaults to [Comparable.compare].
+   */
+  factory PriorityQueue([int comparison(E e1, E e2)]) = HeapPriorityQueue<E>;
+
+  /**
    * Number of elements in the queue.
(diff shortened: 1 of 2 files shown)


lib/priority_queue.dart
```diff
@@ -13,6 +13,16 @@ import "dart:collection" show SplayTreeSet;
...
 abstract class PriorityQueue<E> {
   /**
+   * Creates an empty [PriorityQueue].
+   *
+   * The created [PriorityQueue] is a plain [HeapPriorityQueue]. As such,
+   * [Comparable.compare] is used to compare the priority of elements.
+   *
+   * The queue is equivalent to one created by `new HeapPriorityQueue<E>()`.
+   */
+  factory PriorityQueue() = HeapPriorityQueue<E>;
```
nex3: It would be good to give this a comparator argument as well.

a14n: I used the same pattern as `Set()` that forward to `LinkedHashSet` (https://github.com/dart-lang/sdk/blob/master/sdk/lib/core/set.dart#L45 and https://github.com/dart-lang/sdk/blob/master/sdk/lib/collection/linked_hash_set.dart#L77). Let me know what you prefer.

nex3: I'm not really sure why the default Set constructor doesn't have those arguments, actually. Regardless, the ordering of a priority queue is pretty central to its behavior, so it should be customizable in the main constructor.

a14n: done (commit amended)

PR #22, opened 2016-03-07
Rename toString in wrapper_test

### LINKED ISSUE #21: wrapper_test has analyzer warnings
in `wrapper_test.dart`, I see the following with the analyzer:

```
177: [INFO] The getter 'equals' is not defined for the class 'String' [UNDEFINED_GETTER]
263: [INFO] The getter 'equals' is not defined for the class 'String' [UNDEFINED_GETTER]
```

These are because `Expector` is overriding `Object`'s `toString()` with a method that doesn't return a String. Should we change `Expector`'s method to `asString()` or something similar?

srawlins:
Fixes #21. Tests still pass.

### DIFF
--- a/test/wrapper_test.dart
+++ b/test/wrapper_test.dart
@@ -36,7 +36,9 @@ abstract class Expector {
     testInvocations(m, m2);
   }));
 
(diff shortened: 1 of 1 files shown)


nex3:
Thanks!

PR #26, opened 2016-03-26
Fix types for binarySearch and lowerBound

abarth:
The arguments to these functions need not be Comparable. The caller can
supply a compare function that is used to compare the elements of the
list in the event that they don't implement Comparable.

### DIFF
--- a/lib/src/algorithms.dart
+++ b/lib/src/algorithms.dart
@@ -32,10 +32,10 @@ int _comparableBinarySearch/*<T extends Comparable<T>>*/(
 /// the objects.
 ///
 /// Returns -1 if [value] is not in the list by default.
-int binarySearch/*<T extends Comparable<T>>*/(
+int binarySearch/*<T>*/(
     List/*<T>*/ sortedList, /*=T*/ value, { int compare(/*=T*/ a, /*=T*/ b) }) {
   if (compare == null) {
-    return _comparableBinarySearch(sortedList, value);
+    return _comparableBinarySearch(sortedList as List<Comparable>, value as Comparable);
   }
   int min = 0;
   int max = sortedList.length;
@@ -81,10 +81,10 @@ int _comparableLowerBound(List<Comparable> list, Comparable value) {
 ///
 /// Returns [sortedList.length] if all the items in [sortedList] compare less
 /// than [value].
-int lowerBound/*<T extends Comparable<T>>*/(
+int lowerBound/*<T>*/(
     List/*<T>*/ sortedList, /*=T*/ value, { int compare(/*=T*/ a, /*=T*/ b) }) {
   if (compare == null) {
-    return _comparableLowerBound(sortedList, value);
+    return _comparableLowerBound(sortedList as List<Comparable>, value as Comparable);
   }
   int min = 0;
   int max = sortedList.length;


abarth:
@nex3 

These types were added in https://codereview.chromium.org/1831103004, which broke Flutter because we call these functions with non-Comparable arguments and supply a compare function.

lib/src/algorithms.dart
```diff
@@ -32,10 +32,10 @@ int _comparableBinarySearch/*<T extends Comparable<T>>*/(
 /// the objects.
 ///
 /// Returns -1 if [value] is not in the list by default.
-int binarySearch/*<T extends Comparable<T>>*/(
+int binarySearch/*<T>*/(
     List/*<T>*/ sortedList, /*=T*/ value, { int compare(/*=T*/ a, /*=T*/ b) }) {
   if (compare == null) {
-    return _comparableBinarySearch(sortedList, value);
+    return _comparableBinarySearch(sortedList as List<Comparable>, value as Comparable);
```
Hixie: should use an assignment cast here target than `as`

abarth: That won't pass the strong mode analyzer that package:collection uses.

nex3: `as List<Comparable>` will break here if the list doesn't have a reified runtime type of `Comparable`, which we can't count on. What we've done in [`PriorityQueue`](https://github.com/dart-lang/collection/blob/master/lib/src/priority_queue.dart#L146-L147) is to have the `compare()` function default to a version that does an internal `as` check before calling `compareTo()`. The simplest solution would be to do that here and just delete `_comparableBinarySearch()` entirely.

Also, we should more clearly document that this will throw a runtime error if `compare()` is omitted and `T` isn't a subtype of `Comparable<T>`.

abarth: > Also, we should more clearly document that this will throw a runtime error if compare() is omitted and T isn't a subtype of Comparable<T>.

Why isn't that sufficient?  Are you worried that the List will fail `as List<Comparable>` because it might be a `List<Object>` but that it will just-so-happen that all the elements of the list will actually pass `as Comparable` ?

nex3: Yes. That's very easy right now—users who aren't using strong mode have no guarantees that their reified generics are being propagated as they expect, and even if they did the VM wouldn't do it for generic methods.

Hixie: List<dynamic> will be as List<Comparable>. How could you get from a List<Comparable> to something that isn't List<Comparable>?

nex3: `List<dynamic>` isn't `as`-compatible with `List<Comparable>` in strong mode.

Hixie: In what sense? Isn't strong mode purely an analyzer feature?

The analyzer does say "`x (List<dynamic>) will need runtime check to cast to type List<Comparable>`", but a runtime check is fine here, no?

nex3: There's a runtime component to strong mode too. It's currently only implemented by DDC, but I believe there are plans to expand it to the VM and dart2js as well. Among other things, the runtime component brings the static strong mode's notion of type compatibility to `is` and `as` checks.

Hixie: Isn't the strong mode's notion of type compatibility "that will require a runtime check"? (Which would be fine here?)

I guess I don't understand the problem. Is this documented somewhere so I can read up on it?

nex3: > Isn't the strong mode's notion of type compatibility "that will require a runtime check"? (Which would be fine here?)

At runtime, `is` and `as` checks against generics in strong mode require that the reified type of the object matches the specified generic type. So `[] as List<String>` will work in dart2js but fail in DDC.

> I guess I don't understand the problem. Is this documented somewhere so I can read up on it?

https://github.com/dart-lang/dev_compiler/blob/master/doc/RUNTIME_SAFETY.md

leafpetersen: Just a drive by comment:  you guys (flutter) are mostly using strong mode just for the static checking, but packages like this will also be used with DDC, which injects a runtime checking component.  So we want to be sure that any `as` checks that are inserted are compatible with the DDC runtime checks in the expected use cases.  I don't know what contract is expected of this function: but, if it is part of the contract that `binarySearch<dynamic>(<dynamic>[xxxx], x)` should work, where `xxx` is a series of things that implement `Comparable`.... then that cast is wrong, since on DDC it will fail.  If the contract is that only lists which were created as `List<Comparable>` (or subtypes) are allowed, then the cast is ok.

abarth: Yes, I understand.  I just ran into this issue when we updated package:collection and the analyzer turned red because we were calling this function with a List<Foo> where Foo didn't implement Comparable.  Previously that was permitted because this function takes a `compare` argument to compare object that don't implement `Comparable`.  Currently that's disallowed by the analyzer.  I'm not sure whether changes that break only the analyzer count as breaking changes for semantic versioning, but they turn our buildbot red.

I'm not sure what the intent of the package:collection authors is here.  I suspect it might be time for me to step back and let @nex3 or whoever decide what they think is best.

I guess I could file an issue about how package:collection version 1.4.1 breaks analyzer compatbility with version 1.4.0, but there isn't really anything that can be done about that issue anymore because version 1.4.1 has been published to pub and cannot be rolled back.

leafpetersen: I think we (or at least I) have been very focused on avoiding changes that are breaking with respect to standard (spec-mode) users and haven't been giving sufficient thought to breaking changes with respect to strong mode users.  I'll bring this up with the team.

abarth: Does the analyzer consider generic method annotations only when run in strong mode?

abarth:
Closing this pull request to turn this issue over to @nex3

lib/src/algorithms.dart
leafpetersen: At the moment, yes.  The comment-based generic method syntax is only turned on by default in the analyzer when in strong mode (it may be possible to turn it on in spec-mode, not sure).  The analyzer also supports the actual proposed syntax (basically remove the /**/ from the comment based syntax) behind a flag.  Support for generic methods in other tools (dart2js, VM) is further behind, hence the comment based syntax.

abarth: Thanks for the background.

PR #32, opened 2016-08-16
fix docs

mockturtl:
External links render as code, for example: https://www.dartdocs.org/documentation/collection/1.9.0/collection/compareNatural.html


nex3:
Thanks for the patch!

Issue #38, opened 2016-11-13
Proposal: Equality<E>.from(<closure>)


matanlurey:
There should be a less verbose way to create an `Equality` implementation for:

```dart
abstract class Employee {
  int get idNumber;
  String get name;
}

void main() {
  final equality = new Equality<Employee>.from((e) => e.idNumber);
  assert(
    equality.equals(new Employee(idNumber: 5), new Employee(idNumber: 5)),
  );
}
```

An implementation example:

```dart
typedef Object _GetKey<E>(E object);

class _ClosureEquality<E> implements Equality<E> {
  final _GetKey<E> _getKey;

  _ClosureEquality(Object getKey(E object)) : this._getKey = getKey;

  @override
  bool isValidKey(Object o) => o is E;

  @override
  int hash(E e) => _getKey(e).hashCode;

  @override
  bool equals(E e1, E e2) => _getKey(e1) == _getKey(e2);
}
```

nex3:
I like the idea, but I'm not sure I like the name. Without further context, I'd expect `Equality.from()` to take closures for `equals()` and `hash()`. What about `Equality.by()`, like you might say "the employees are equal by ID?"

It would also be good to include an optional parameter to define how the closure results are compared. Unfortunately factory constructors can't have method-level generics, so you'd have to define a new class to type this properly:

```dart
class EqualityBy<E, F> implements Equality<E> {
  EqualityBy(F getKey(E object), [Equality<F> inner]);
}
```

matanlurey:
I admit I am the _worst_ at naming. I like `Equality.by()` 👍 

If you're OK with the API / changes, I'll start on a PR.

nex3:
Go for it.

PR #40, opened 2016-11-15
Add EqualityBy

matanlurey:
Ran dartfmt on modified files.

Closes https://github.com/dart-lang/collection/issues/38

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 1.13.0
+
+* Add `EqualityBy`
+
 ## 1.12.1
 
 * Fix bug in `equalsIgnoreAsciiCase`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,6 @@
 
 * Add `EqualityBy`
 
-## 1.12.1
-
-* Fix bug in `equalsIgnoreAsciiCase`.
-
 ## 1.12.0
 
 * Add `CaseInsensitiveEquality`.
--- a/lib/src/equality.dart
+++ b/lib/src/equality.dart
@@ -35,16 +35,16 @@ typedef F _GetKey<E, F>(E object);
 /// Equality of objects based on derived values.
 ///
 /// For example, given the class:
-///     ```
-///     abstract class Employee {
-///       int get employmentId;
-///     }
-///     ```
+/// ```dart
+/// abstract class Employee {
+///   int get employmentId;
+/// }
+/// ```
 ///
 /// The following [Equality] considers employees with the same IDs to be equal:
-///     ```
-///     new EqualityBy((Employee e) => e.employmentId);
-///     ```
+/// ```dart
+/// new EqualityBy((Employee e) => e.employmentId);
+/// ```
 ///
 /// It's also possible to pass an additional equality instance that should be
 /// used to compare the value itself.
--- a/lib/src/equality.dart
+++ b/lib/src/equality.dart
@@ -32,34 +32,44 @@ abstract class Equality<E> {
 
 typedef F _GetKey<E, F>(E object);
 
-/// Equality of objects using a canonical value of the object itself.
+/// Equality of objects based on derived values.
 ///
 /// For example, given the class:
+///     ```
 ///     abstract class Employee {
 ///       int get employmentId;
 ///     }
+///     ```
 ///
-/// Creating an [Equality] object that uses `employmentId`:
-///     new EqualityBy<Employee, int>((e) => e.employmentId);
+/// The following [Equality] considers employees with the same IDs to be equal:
+///     ```
+///     new EqualityBy((Employee e) => e.employmentId);
+///     ```
 ///
 /// It's also possible to pass an additional equality instance that should be
 /// used to compare the value itself.
 class EqualityBy<E, F> implements Equality<E> {
+  // Returns a derived value F from an object E.
   final _GetKey<E, F> _getKey;
+
+  // Determines equality between two values of F.
   final Equality<F> _inner;
 
   EqualityBy(F getKey(E object), [Equality<F> inner = const DefaultEquality()])
       : _getKey = getKey,
         _inner = inner;
 
-  @override
   bool equals(E e1, E e2) => _inner.equals(_getKey(e1), _getKey(e2));
 
-  @override
   int hash(E e) => _inner.hash(_getKey(e));
 
-  @override
-  bool isValidKey(Object o) => o is E;
+  bool isValidKey(Object o) {
+    if (o is E) {
+      final value = _getKey(o);
+      return value is F && _inner.isValidKey(value);
+    }
+    return false;
+  }
 }
 
 /// Equality of objects that compares only the natural equality of the objects.
--- a/lib/src/equality.dart
+++ b/lib/src/equality.dart
@@ -30,6 +30,38 @@ abstract class Equality<E> {
   bool isValidKey(Object o);
 }
 
+typedef F _GetKey<E, F>(E object);
+
+/// Equality of objects using a canonical value of the object itself.
+///
+/// For example, given the class:
+///     abstract class Employee {
+///       int get employmentId;
+///     }
+///
+/// Creating an [Equality] object that uses `employmentId`:
+///     new EqualityBy<Employee, int>((e) => e.employmentId);
+///
+/// It's also possible to pass an additional equality instance that should be
+/// used to compare the value itself.
+class EqualityBy<E, F> implements Equality<E> {
+  final _GetKey<E, F> _getKey;
+  final Equality<F> _inner;
+
+  EqualityBy(F getKey(E object), [Equality<F> inner = const DefaultEquality()])
+      : _getKey = getKey,
+        _inner = inner;
+
+  @override
+  bool equals(E e1, E e2) => _inner.equals(_getKey(e1), _getKey(e2));
+
+  @override
+  int hash(E e) => _inner.hash(_getKey(e));
+
+  @override
+  bool isValidKey(Object o) => o is E;
+}
+
 /// Equality of objects that compares only the natural equality of the objects.
 ///
 /// This equality uses the objects' own [Object.==] and [Object.hashCode] for
--- a/test/equality_test.dart
+++ b/test/equality_test.dart
@@ -234,9 +234,3 @@ class Element implements Comparable<Element> {
   bool operator==(Object other) => other is Element && id == other.id;
   int compareTo(other) => id.compareTo(other.id);
 }
-
-/// Wrapper objects for an `id` value without implementing proper equality.
-class NonCanonicalElement {
-  final String id;
-  NonCanonicalElement(this.id);
-}
--- a/test/equality_test.dart
+++ b/test/equality_test.dart
@@ -191,11 +191,11 @@ main() {
           equality1.equals(
               ["foo", "foo"], ["bar", "bar"]),
           isFalse);
-
     });
 
     test("equality with an inner equality", () {
       expect(equality2.equals(["fOo"], ["FoO"]), isTrue);
+      expect(equality2.equals(["foo"], ["ffõõ"]), isFalse);
     });
 
     test("hash", () {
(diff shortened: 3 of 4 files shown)


nex3:
Please don't dartfmt this package.

matanlurey:
> Please don't dartfmt this package.

@nex3 Any particular reason? It makes small changes really difficult - I either have to just try to oddly get them to fit the current style, or just not care at all (?). Seems inconsistent with most of the other dart-lang packages as well.

nex3:
`dartfmt` still handles some things really poorly, including multiline string literals and very large collection literals. I also find that it rarely formats better than I do, so packages I maintain are formatted by hand.

matanlurey:
Are you the only author though - it looks like there is 12 total contributors:
https://github.com/dart-lang/collection/graphs/contributors

How does everyone know how to format it like @nex3? I certainly don't.

nex3:
Other than following the style guide, I'm not that stringent. Is there something in particular you aren't sure how to format?

matanlurey:
OK, I reverted dartfmt on anything I didn't personally write in this PR.

PTAL.

lib/src/equality.dart
```diff
@@ -30,6 +30,38 @@ abstract class Equality<E> {
...
+
+/// Equality of objects using a canonical value of the object itself.
+///
+/// For example, given the class:
+///     abstract class Employee {
+///       int get employmentId;
+///     }
+///
+/// Creating an [Equality] object that uses `employmentId`:
+///     new EqualityBy<Employee, int>((e) => e.employmentId);
```
nex3: Maybe write this as `new EqualityBy((Employee e) => e.employmentId)`? Normally we don't type-annotate function parameters, but it's terser than explicitly annotating both type parameters.

lib/src/equality.dart
```diff
@@ -30,6 +30,38 @@ abstract class Equality<E> {
...
+typedef F _GetKey<E, F>(E object);
+
+/// Equality of objects using a canonical value of the object itself.
+///
+/// For example, given the class:
+///     abstract class Employee {
+///       int get employmentId;
+///     }
+///
+/// Creating an [Equality] object that uses `employmentId`:
```
nex3: Be a little more explicit. "The following [Equality] considers employees with the same IDs to be equal:"

lib/src/equality.dart
```diff
@@ -30,6 +30,38 @@ abstract class Equality<E> {
...
+///     new EqualityBy<Employee, int>((e) => e.employmentId);
+///
+/// It's also possible to pass an additional equality instance that should be
+/// used to compare the value itself.
+class EqualityBy<E, F> implements Equality<E> {
+  final _GetKey<E, F> _getKey;
+  final Equality<F> _inner;
```
nex3: Document these.

lib/src/equality.dart
```diff
@@ -30,6 +30,38 @@ abstract class Equality<E> {
...
+
+  EqualityBy(F getKey(E object), [Equality<F> inner = const DefaultEquality()])
+      : _getKey = getKey,
+        _inner = inner;
+
+  @override
```
nex3: This package doesn't use `@override`.

lib/src/equality.dart
```diff
@@ -30,6 +30,38 @@ abstract class Equality<E> {
...
+  bool equals(E e1, E e2) => _inner.equals(_getKey(e1), _getKey(e2));
+
+  @override
+  int hash(E e) => _inner.hash(_getKey(e));
+
+  @override
+  bool isValidKey(Object o) => o is E;
```
nex3: Should this also check `_inner.isValidKey(_getKey(o))`?

test/equality_test.dart
```diff
@@ -175,6 +175,25 @@ main() {
     expect(equality.hash("foo"), isNot(equals(equality.hash("bar"))));
     expect(equality.hash("fÕÕ"), isNot(equals(equality.hash("fõõ"))));
   });
+
+  test("EqualityBy", () {
+    var equality = new EqualityBy<NonCanonicalElement, String>((e) => e.id);
```
nex3: I'd probably avoid defining a custom class here and just have this be `new EqualityBy<List<String>, String>((e) => e.first)` (or `new EqualityBy((List<String> e) => e.first)`).

test/equality_test.dart
```diff
@@ -175,6 +175,25 @@ main() {
...
+        equality.equals(
+            new NonCanonicalElement("foo"), new NonCanonicalElement("foo")),
+        isTrue);
+    expect(
+        equality.equals(
+            new NonCanonicalElement("foo"), new NonCanonicalElement("bar")),
+        isFalse);
+
+    equality = new EqualityBy<NonCanonicalElement, String>(
+        (e) => e.id, const CaseInsensitiveEquality());
```
nex3: Make this a separate test in the same group.

test/equality_test.dart
```diff
@@ -175,6 +175,25 @@ main() {
...
+  test("EqualityBy", () {
+    var equality = new EqualityBy<NonCanonicalElement, String>((e) => e.id);
+    expect(
+        equality.equals(
+            new NonCanonicalElement("foo"), new NonCanonicalElement("foo")),
+        isTrue);
+    expect(
+        equality.equals(
+            new NonCanonicalElement("foo"), new NonCanonicalElement("bar")),
+        isFalse);
```
nex3: Test `hash()` and `isValidKey()` as well (probably as separate tests in the same group.

test/equality_test.dart
```diff
@@ -175,6 +175,25 @@ main() {
...
+
+    equality = new EqualityBy<NonCanonicalElement, String>(
+        (e) => e.id, const CaseInsensitiveEquality());
+    expect(
+        equality.equals(
+            new NonCanonicalElement("fOo"), new NonCanonicalElement("FoO")),
+        isTrue);
```
nex3: Test a case where this equality fails as well.

CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+## 1.13.0
+
+* Add `EqualityBy`
+
+## 1.12.1
+
+* Fix bug in `equalsIgnoreAsciiCase`.
```
matanlurey: Done

lib/src/equality.dart
```diff
@@ -30,6 +30,38 @@ abstract class Equality<E> {
   bool isValidKey(Object o);
 }
 
+typedef F _GetKey<E, F>(E object);
+
+/// Equality of objects using a canonical value of the object itself.
```
matanlurey: Done.

lib/src/equality.dart
```diff
@@ -30,6 +30,38 @@ abstract class Equality<E> {
...
+///
+/// For example, given the class:
+///     abstract class Employee {
```
matanlurey: Done.

lib/src/equality.dart
matanlurey: Done.

lib/src/equality.dart
matanlurey: Done.

lib/src/equality.dart
matanlurey: Done.

lib/src/equality.dart
matanlurey: Done.

lib/src/equality.dart
matanlurey: Done.

test/equality_test.dart
matanlurey: Done.

test/equality_test.dart
matanlurey: Done.

lib/src/equality.dart
```diff
@@ -32,34 +32,44 @@ abstract class Equality<E> {
...
 
-/// Equality of objects using a canonical value of the object itself.
+/// Equality of objects based on derived values.
 ///
 /// For example, given the class:
+///     ```
 ///     abstract class Employee {
 ///       int get employmentId;
 ///     }
+///     ```
```
nex3: If you're using triple-tics, you don't need to indent. Also, add `dart` after the first set so that we get syntax highlighting:

    ```dart
    code
    ````

lib/src/equality.dart
```diff
@@ -32,34 +32,44 @@ abstract class Equality<E> {
...
 
-  @override
   int hash(E e) => _inner.hash(_getKey(e));
 
-  @override
-  bool isValidKey(Object o) => o is E;
+  bool isValidKey(Object o) {
+    if (o is E) {
+      final value = _getKey(o);
+      return value is F && _inner.isValidKey(value);
```
nex3: I don't think this `is F` check is necessary, since `_getKey()` is typed to return an `F`. As a general rule, I don't write runtime checks that are unnecessary if user code is strong-mode clean.

test/equality_test.dart
```diff
@@ -176,23 +176,42 @@ main() {
...
+    test("equality", () {
+      expect(
+          equality1.equals(
+              ["foo", "foo"], ["foo", "bar"]),
+          isTrue);
+      expect(
+          equality1.equals(
+              ["foo", "foo"], ["bar", "bar"]),
+          isFalse);
+
```
nex3: Nit: unnecessary newline.

test/equality_test.dart
```diff
@@ -176,23 +176,42 @@ main() {
...
+    });
+
+    test("equality with an inner equality", () {
+      expect(equality2.equals(["fOo"], ["FoO"]), isTrue);
```
nex3: Add a negative case here as well.

test/equality_test.dart
```diff
@@ -176,23 +176,42 @@ main() {
...
+
+    test("hash with an inner equality", () {
+      expect(
+          equality2.hash(["fOo"]), const CaseInsensitiveEquality().hash("foo"));
+    });
+
+    test("isValidKey", () {
+      expect(equality1.isValidKey(["foo"]), isTrue);
+      expect(equality1.isValidKey("foo"), isFalse);
+      expect(equality1.isValidKey([1]), isFalse);
```
nex3: I'm guessing this is supposed to test that `EqualityBy.isValidKey()` calls `CaseSensitiveEquality.isValidKey()`, but it won't do that in a strong-mode runtime. `[1] is F` will fail, so the inner call will never happen. You'd need to create a new EqualityBy with Object type parameters.

lib/src/equality.dart
matanlurey: It's required for strong mode unfortunately, otherwise I am calling `_getKey(Object)`, which is a warning.

test/equality_test.dart
matanlurey: Done

test/equality_test.dart
matanlurey: Done

test/equality_test.dart
matanlurey: I had to keep the `if (o is E) {`, so this test validates it.

nex3: Got it. It would still be good to have a case that *does* hit the inner `Equality.isValidKey()`, though.

matanlurey: Done.

test/equality_test.dart
```diff
@@ -175,6 +175,51 @@ main() {
...
     expect(equality.hash("fÕÕ"), isNot(equals(equality.hash("fõõ"))));
   });
+
+  group("EqualityBy should use a derived value for ", () {
+    final equality1 = new EqualityBy<List<String>, String>(
+        (e) => e.first);
+    final equality2 = new EqualityBy<List<String>, String>(
+        (e) => e.first, const CaseInsensitiveEquality());
+    final equality3 = new EqualityBy<List<Object>, Object>(
+        (e) => e.first, const IterableEquality());
```
matanlurey: I'll make this change, but in the future it's a little unreasonable to leave this nitpick after a month of PR back and forth. It definitely doesn't make it easy to encourage contributions from others.

matanlurey: Updated PR.

nex3:
Released as 1.13.0.

PR #41, opened 2016-11-15
Resolve some strong-mode warnings encountered in IntelliJ

matanlurey:
It looks like type inference still needs some manual help, so I helped where it was simple to do.

### DIFF
--- a/test/comparators_test.dart
+++ b/test/comparators_test.dart
@@ -85,7 +85,7 @@ void main() {
   // This will sort alphabetically (by charcode) the way digits sort
   // numerically, and the leading 0 means it sorts like a digit
   // compared to non-digits.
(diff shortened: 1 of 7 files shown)


test/comparators_test.dart
```diff
@@ -85,7 +85,7 @@ void main() {
   // This will sort alphabetically (by charcode) the way digits sort
   // numerically, and the leading 0 means it sorts like a digit
   // compared to non-digits.
-  replaceNumbers(string) => string.replaceAllMapped(new RegExp(r"\d+"), (m) {
+  replaceNumbers(String string) => string.replaceAllMapped(new RegExp(r"\d+"), (m) {
```
nex3: Long line.

nex3: Sorry, this is probably just my GitHub display being too narrow.

Issue #7, opened 2015-06-04
Add UnmodifiableListMixin to pkg/collection


DartBot:
<a href="https://github.com/nex3"><img src="https://avatars.githubusercontent.com/u/188?v=3" align="left" width="96" height="96"hspace="10"></img></a> **Issue by [nex3](https://github.com/nex3)**
_Originally opened as dart-lang/sdk#18970_

---

UnmodifiableMapMixin exists, so it seems strange that UnmodifiableListMixin doesn't as well.

See also issue dart-lang/sdk#11178.

MichaelFenwick:
I too would like this to be added. I'm trying to create an unmodifiable custom collection class and find myself having to manually reimplement everything that UnmodifiableListMixin already does. I'll also add that i would like to have UnmodifiableListBase available in the dart:collection package, but as it is easy to recreate from the ListBase and UnmodifiableListMixin, it is of lower importance to me.

matanlurey:
This has been open since 2015 without action so I am closing as stale.

PR #50, opened 2017-01-20
Add CombinedListView

matanlurey:
Partially addresses https://github.com/dart-lang/core/issues/711


lib/src/combined_wrappers.dart
```diff
@@ -0,0 +1,73 @@
...
+import 'dart:collection';
+
+/// Returns a new list that represents [lists] flattened into a single list.
+///
+/// All methods and accessors treat the new list as-if it were a single
+/// concatenated list, but the underlying implementation is based on lazily
+/// accessing individual list instances.
+///
+/// The returned list is unmodifiable.
+List/*<T>*/ combineLists/*<T>*/(List<List/*<T>*/> lists) {
```
nex3: My intuition is that this should be exposed as a class, to better express that you're constructing a new view over objects rather than just concatenating lists. I'd also call the class `CombinedListView` to emphasize the view aspect further.

lib/src/combined_wrappers.dart
```diff
@@ -0,0 +1,73 @@
...
+
+import 'dart:collection';
+
+/// Returns a new list that represents [lists] flattened into a single list.
+///
+/// All methods and accessors treat the new list as-if it were a single
+/// concatenated list, but the underlying implementation is based on lazily
+/// accessing individual list instances.
+///
+/// The returned list is unmodifiable.
```
nex3: This should be explicit that indexing and `length` are `O(lists)` rather than `O(1)`, and that the resulting list is unmodifiable.

lib/src/combined_wrappers.dart
```diff
@@ -0,0 +1,73 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
```
nex3: I'd prefer this file just be `combined_list.dart`, possibly in a `combined_wrappers` directory.

lib/src/combined_wrappers.dart
```diff
@@ -0,0 +1,73 @@
...
+/// All methods and accessors treat the new list as-if it were a single
+/// concatenated list, but the underlying implementation is based on lazily
+/// accessing individual list instances.
+///
+/// The returned list is unmodifiable.
+List/*<T>*/ combineLists/*<T>*/(List<List/*<T>*/> lists) {
+  return new _CombinedList/*<T>*/(lists);
+}
+
+class _CombinedList<T> extends ListBase<T> implements UnmodifiableListView<T> {
```
nex3: Why not `with UnmodifiableListMixin`?

lib/src/combined_wrappers.dart
```diff
@@ -0,0 +1,73 @@
...
+  static void _throw() {
+    throw new UnsupportedError('Cannot modify an unmodifiable List');
+  }
+
+  final List<List<T>> _lists;
+
+  _CombinedList(this._lists);
+
+  @override
```
nex3: This package doesn't use `@override`.

lib/src/combined_wrappers.dart
```diff
@@ -0,0 +1,73 @@
...
+  set length(int length) {
+    _throw();
+  }
+
+  @override
+  int get length => _lists.fold(0, (length, list) => length + list.length);
+
+  @override
+  T operator [](int index) {
+    final initialIndex = index;
```
nex3: `var`, here and elsewhere

lib/src/combined_wrappers.dart
```diff
@@ -0,0 +1,73 @@
...
+    for (var i = 0; i < _lists.length; i++) {
```
nex3: Why not `var list in _lists`?

lib/src/combined_wrappers.dart
```diff
@@ -0,0 +1,73 @@
...
+      var list = _lists[i];
+      if (index < list.length) {
+        return list[index];
+      }
+      index -= list.length;
+    }
+    throw new RangeError.value(initialIndex, 'index', 'Length of $length');
```
nex3: This should probably be [`new RangeError.index()`](https://api.dartlang.org/stable/1.21.1/dart-core/RangeError/RangeError.index.html).

test/combined_wrappers_test.dart
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) 2013, the Dart project authors.  Please see the AUTHORS file
```
nex3: 2017

test/combined_wrappers_test.dart
```diff
@@ -0,0 +1,18 @@
...
+
+void main() {
+  final list1 = const [1, 2, 3];
+  final list2 = const [4, 5, 6];
+  final list3 = const [7, 8, 9];
+  final concat = []..addAll(list1)..addAll(list2)..addAll(list3);
+  final combine = combineLists([list1, list2, list3]);
+
+  // In every way possible this should test the same as an UnmodifiableListView.
+  common.testUnmodifiableList(concat, combine, 'combineLists');
```
nex3: Also test an empty source list and a source list containing empty lists.

nex3 (changes requested):

lib/src/combined_wrappers.dart
matanlurey: Isn't the fact that it is a `CombinedListView` an implementation detail?

From a maintenance perspective you're adding an interface that others could implement/extend or reference even though there isn't any actual real use for doing so.

lib/src/combined_wrappers.dart
matanlurey: `UnmodifiableListMixin` is part of the SDK and not public :-/

https://github.com/dart-lang/collection/issues/7

lib/src/combined_wrappers.dart
matanlurey: Mostly dart2js-related paranoia (no reason to allocate an iterator)

test/combined_wrappers_test.dart
matanlurey: Done.

test/combined_wrappers_test.dart
matanlurey: Done.

lib/src/combined_wrappers.dart
matanlurey: Done.

lib/src/combined_wrappers.dart
matanlurey: Done.

lib/src/combined_wrappers.dart
matanlurey: Done.

lib/src/combined_wrappers.dart
matanlurey: Done.

lib/src/combined_wrappers.dart
matanlurey: Done.

lib/src/combined_wrappers/combined_list.dart
```diff
@@ -10,8 +10,13 @@ import 'dart:collection';
 /// concatenated list, but the underlying implementation is based on lazily
 /// accessing individual list instances.
 ///
-/// The returned list is unmodifiable.
+/// The resulting list has an index operator (`[]`) and `length` property that
+/// are both `O(lists)`, rather than `O(1)`, and the list is unmodifiable.
 List/*<T>*/ combineLists/*<T>*/(List<List/*<T>*/> lists) {
+  // Small optimization when there are no lists to avoid allocation.
+  if (lists.isEmpty) {
```
nex3: Maybe `lists.every((list) => list.isEmpty)`?

lib/src/combined_wrappers.dart
nex3: > Isn't the fact that it is a `CombinedListView` an implementation detail?

I think a class better expresses what's actually happening. Functions generally produce outputs that are, after the function is finished running, largely independent of their arguments. Classes are much more likely to retain a reference to their constructor argument and update their state based on changes, and the word "View" explicitly indicates the way that will happen.

It's also more consistent with similar existing patterns—we have `Unmodifiable*View` classes, not `make*Unmodifiable()` functions.

> I could have it return an `UnmodifiableListView<T>` though.

*Hopefully* no one is relying on that static type anywhere.

lib/src/combined_wrappers/combined_list.dart
matanlurey: Done, though without the closure allocation.

Also added an optimization for a single list, which is probably common enough too.

test/combined_wrappers_test.dart
nex3: I'd like to also test a source list containing empty lists, but not *only* empty lists.

nex3 (changes requested):

lib/src/combined_wrappers/combined_list.dart
```diff
@@ -0,0 +1,70 @@
...
+/// All methods and accessors treat the new list as-if it were a single
+/// concatenated list, but the underlying implementation is based on lazily
+/// accessing individual list instances.
+///
+/// The resulting list has an index operator (`[]`) and `length` property that
+/// are both `O(lists)`, rather than `O(1)`, and the list is unmodifiable.
+List/*<T>*/ combineLists/*<T>*/(List<List/*<T>*/> lists) {
+  // Small optimization when there are no lists to avoid allocation.
+  if (lists.isEmpty) {
+    return const [];
```
matanlurey: Hmm. I agree with this. Changed it up so the underlying list can always be changed.

lib/src/combined_wrappers/combined_list.dart
```diff
@@ -0,0 +1,70 @@
...
+import 'dart:collection';
+
+/// Returns a new list that represents [lists] flattened into a single list.
+///
+/// All methods and accessors treat the new list as-if it were a single
+/// concatenated list, but the underlying implementation is based on lazily
+/// accessing individual list instances.
+///
+/// The resulting list has an index operator (`[]`) and `length` property that
+/// are both `O(lists)`, rather than `O(1)`, and the list is unmodifiable.
```
matanlurey: Done.

matanlurey:
Friendly ping @nex3

nex3:
Oh, I didn't realize this was ready for review—[this comment](https://github.com/dart-lang/collection/pull/50#discussion_r98311210) is still outstanding.

lib/src/combined_wrappers.dart
matanlurey: Alright. I still don't really agree with this change, but willing to overlook it.

PTAL.

lib/src/combined_wrappers/combined_list.dart
```diff
@@ -23,17 +23,17 @@ List/*<T>*/ combineLists/*<T>*/(List<List/*<T>*/> lists) {
   if (lists.length == 1) {
     return new UnmodifiableListView/*<T>*/(lists.first);
   }
-  return new _CombinedList/*<T>*/(lists);
+  return new CombinedList/*<T>*/._(lists);
 }
 
-class _CombinedList<T> extends ListBase<T> implements UnmodifiableListView<T> {
+class CombinedList<T> extends ListBase<T> implements UnmodifiableListView<T> {
```
nex3: `CombinedListView`

Also, remove `combinedList()` in favor of a `CombinedListView` constructor.

lib/src/combined_wrappers/combined_list.dart
```diff
@@ -10,63 +10,66 @@ import 'dart:collection';
...
-/// The returned list is unmodifiable.
+/// The resulting list has an index operator (`[]`) and `length` property that
+/// are both `O(lists)`, rather than `O(1)`, and the list is unmodifiable - but
+/// underlying changes to these lists are still accessible from the resulting
+/// list.
 List/*<T>*/ combineLists/*<T>*/(List<List/*<T>*/> lists) {
-  return new _CombinedList/*<T>*/(lists);
+  // Small optimization when there are no lists to avoid allocation.
+  if (lists.isEmpty) {
+    return const [];
```
nex3: This still behaves inconsistently when `lists` itself is modified. I think we should just construct a `CombinedList()` regardless of the size of the input, and allow `lists` to change after the fact just like its contents.

nex3 (changes requested):

lib/src/combined_wrappers/combined_list.dart
matanlurey: Done.

lib/src/combined_wrappers/combined_list.dart
matanlurey: Done.

nex3 (commented):
Please add a test for modifying the `List<List>`—then this'll be good to go!

matanlurey:
@nex3 Added the test case!

PR #51, opened 2017-01-24
use pkg/test config for test bots

kevmoo:


nex3:
I for one will probably continue not looking at the package bots, but LGTM I guess.

Issue #48, opened 2017-01-20
Add top-level `combine<Iterables|Lists|Maps>`


matanlurey:
Let's start with just discussing `combineList`, which is currently implemented in various forms internally, and has lots of utility and potential perf-saving. 

```dart
//// Returns a new list that represents [lists] flattened into a single list.
///
/// All methods and accessors treat the new list as-if it were a single concatenated list,
/// but the underlying implementation delegates to each corresponding list as necessary. 
List<T> combineLists<T>(List<List<T>> lists) => ...
```

Here is an example of implementing `first`:

```dart
class _CombinedList<T> implements List<T> {
  final List<T> _lists;

  T get first {
    for (final list in _lists) {
     if (list.isNotEmpty) {
       return list.first; 
     }
    }
   throw ...
  }
}
```

/cc @nex3, @leonsenft, @ferhatb 

FWIW I believe @isoos originally implemented something like this for the high-performance table, and it worked really well. I'd also eventually like to see the same thing for `Map` and `Iterable` if possible - though to simplify things they may need to be read-only, but worth discussing anyway.

nex3:
I could see a `CombinedListView` class for this. Does this mean that `[]` would be O(n) for the number of lists that were combined? Or would you track some sort of radix list?

matanlurey:
> I could see a `CombinedListView` class for this. Does this mean that `[]` would be O(n) for the number of lists that were combined? Or would you track some sort of radix list?

I think given that I imagine this is usually just 2 or 3 lists, I'd probably implement as a `O(n)` where _n_ is the number of lists that were combined. That's how it is implemented today and is definitely just a few lines of code.

nex3:
That sounds fine to me.

isoos:
I don't think `O(number of lists|maps combined)` is a big issue. When people want to combine 1000s of lists or maps, they will probably figure out better data structures to represent what they need anyway.

PR #52, opened 2017-02-09
Add CombinedIterableView

matanlurey:
Implements `CombinedIterableView` (https://github.com/dart-lang/collection/issues/48):

* Moves `combined_list_test.dart` -> `test/combined_wrapper/list_test.dart`

Branched from https://github.com/dart-lang/collection/pull/50, I'll merge once committed.

PR #53, opened 2017-02-09
Add CombinedMapView

matanlurey:
Branched from https://github.com/dart-lang/collection/pull/52, which needs to be merged first.

Finally closes https://github.com/dart-lang/collection/issues/48.

### DIFF
--- a/lib/src/combined_wrappers/combined_iterable.dart
+++ b/lib/src/combined_wrappers/combined_iterable.dart
@@ -24,6 +24,8 @@ class CombinedIterableView<T> extends IterableBase<T> {
   // Special cased isEmpty/length since many iterables have an efficient
   // implementation instead of running through the entire iterator.
 
+  bool contains(Object element) => _iterables.any((i) => i.contains(element));
+
   bool get isEmpty => _iterables.every((i) => i.isEmpty);
 
   int get length => _iterables.fold(0, (length, i) => length + i.length);
--- a/lib/src/combined_wrappers/combined_iterable.dart
+++ b/lib/src/combined_wrappers/combined_iterable.dart
@@ -21,8 +21,8 @@ class CombinedIterableView<T> extends IterableBase<T> {
   Iterator<T> get iterator =>
       new _CombinedIterator<T>(_iterables.map((i) => i.iterator).iterator);
 
-  // Special cased isEmpty/length since many iterables have an efficient
-  // implementation instead of running through the entire iterator.
+  // Special cased contains/isEmpty/length since many iterables have an
+  // efficient implementation instead of running through the entire iterator.
 
   bool contains(Object element) => _iterables.any((i) => i.contains(element));
 
--- a/lib/src/combined_wrappers/combined_map.dart
+++ b/lib/src/combined_wrappers/combined_map.dart
@@ -37,5 +37,16 @@ class CombinedMapView<K, V> extends UnmodifiableMapBase<K, V> {
     return null;
   }
 
+  /// The keys of [this].
+  ///
+  /// The returned iterable has efficient `length` and `contains` operations,
+  /// based on [length] and [containsKey] of the individual maps.
+  ///
+  /// The order of iteration is defined by the individual `Map` implementations,
+  /// but must be consistent between changes to the maps.
+  ///
+  /// Unlike most [Map] implementations, modifying an individual map while
+  /// iterating the keys will _sometimes_ throw. This behavior may change in
+  /// the future.
   Iterable<K> get keys => new CombinedIterableView<K>(_maps.map((m) => m.keys));
 }
--- a/lib/src/combined_wrappers/combined_map.dart
+++ b/lib/src/combined_wrappers/combined_map.dart
@@ -6,7 +6,7 @@ import 'dart:collection';
 
 import 'combined_iterable.dart';
 
-/// Returns a new map that represents maps flattened into a single ma[.
+/// Returns a new map that represents maps flattened into a single map.
 ///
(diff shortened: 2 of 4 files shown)


lib/src/combined_wrappers/combined_map.dart
```diff
@@ -0,0 +1,39 @@
...
+/// map.
+class CombinedMapView<K, V> extends UnmodifiableMapBase<K, V> {
+  final Iterable<Map<K, V>> _maps;
+
+  CombinedMapView(this._maps);
+
+  @override
+  V operator [](Object key) {
+    for (var map in _maps) {
+      // Avoid two hash look ups on a positive hit.
```
kevmoo: positive *non-null* hit 😄

matanlurey:
I'll rebase against master and ping when ready.

matanlurey:
Ready for review @nex3!

lib/src/combined_wrappers/combined_map.dart
```diff
@@ -0,0 +1,37 @@
...
+/// multiple maps the first value is returned.
+///
+/// The resulting map has an index operator (`[]`) and `length` property that
+/// are both `O(maps)`, rather than `O(1)`, and the map is unmodifiable - but
+/// underlying changes to these maps are still accessible from the resulting
+/// map.
+class CombinedMapView<K, V> extends UnmodifiableMapBase<K, V> {
+  final Iterable<Map<K, V>> _maps;
+
+  CombinedMapView(this._maps);
```
nex3: Document that the `maps` iterable is accessed repeatedly, so it should be a collection type like `List` or `Set` rather than a lazy iterable produced by `map()` *et al*.

I really wish we had a distinct `Collection` type so we could express that statically.

lib/src/combined_wrappers/combined_map.dart
```diff
@@ -0,0 +1,37 @@
...
+      // Avoid two hash look ups on a positive hit.
+      var value = map[key];
+      if (value != null || map.containsKey(value)) {
+        return value;
+      }
+    }
+    return null;
+  }
+
+  Iterable<K> get keys => new CombinedIterableView<K>(_maps.map((m) => m.keys));
```
nex3: From [the `UnmodifiableMapBase` docs](https://api.dartlang.org/stable/1.22.1/dart-collection/UnmodifiableMapBase-class.html):

> The `keys` iterable should have efficient `length` and `contains` operations, and it should catch concurrent modifications of the keys while iterating.

`CombinedIterableView` doesn't have an efficient `contains()` (maybe that should be special-cased as well). It also won't necessarily throw if a later map is modified while an earlier one is being traversed, but that may not realistically be fixable.

lib/src/combined_wrappers/combined_map.dart
```diff
@@ -0,0 +1,37 @@
...
+/// underlying changes to these maps are still accessible from the resulting
+/// map.
+class CombinedMapView<K, V> extends UnmodifiableMapBase<K, V> {
+  final Iterable<Map<K, V>> _maps;
+
+  CombinedMapView(this._maps);
+
+  V operator [](Object key) {
+    for (var map in _maps) {
+      // Avoid two hash look ups on a positive hit.
```
nex3: Nit: "lookups"

test/combined_wrapper/map_test.dart
```diff
@@ -0,0 +1,55 @@
...
+import 'package:collection/collection.dart';
+import 'package:test/test.dart';
+
+import '../unmodifiable_collection_test.dart' as common;
+
+void main() {
+  final map1 = const {1: 1, 2: 2, 3: 3};
+  final map2 = const {4: 4, 5: 5, 6: 6};
+  final map3 = const {7: 7, 8: 8, 9: 9};
+  final concat = {}..addAll(map1)..addAll(map2)..addAll(map3);
```
nex3: Nit: `var`

lib/src/combined_wrappers/combined_map.dart
matanlurey: I've added a special case for `contains` in `CombinedIterableView`.

I don't think I can do anything about concurrent modifications. I can document that concurrent modifications should be avoided because we don't throw?

test/combined_wrapper/map_test.dart
matanlurey: Done.

lib/src/combined_wrappers/combined_iterable.dart
```diff
@@ -24,6 +24,8 @@ class CombinedIterableView<T> extends IterableBase<T> {
   // Special cased isEmpty/length since many iterables have an efficient
```
nex3: Mention that `contains` is special-cased as well.

lib/src/combined_wrappers/combined_map.dart
nex3: Documenting that would be good. Be sure to mention that it will *sometimes* throw, but maybe don't guarantee the specific circumstances so we can change it in the future if we need to.

matanlurey: Done.

lib/src/combined_wrappers/combined_iterable.dart
matanlurey: Done.

matanlurey:
@kevmoo Do you mind approving? I think I've addressed all of @nex3's feedback but it would be nice not to have to wait several weeks to merge something already vetted :)

matanlurey:
One more tiny ping for @kevmoo

PR #64, opened 2017-11-10
Export classes from dart:collection.

lrhn:
In preparation of moving the classes to this package in Dart 2.0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,18 @@
 ## 1.15.0
 
-* Re-export a number of platform library classes from `dart:collection`
-  in anticipation of moving the classes to this package for Dart 2.0.
+* Re-export the following platform library classes from `dart:collection`
+  in anticipation of moving the classes to this package for Dart 2.0:
+  * `DoubleLinkedQueue`
+  * `DoubleLinkedQueueEntry`
+  * `HasNextIterator`
+  * `LinkedList`
+  * `LinkedListEntry`
+  * `MapView`
+  * `SplayTreeMap`
+  * `SplayTreeSet`
+  * `UnmodifiableListView`
+  * `UnmodifiableMapBase`
+  * `UnmodifiableMapView`
 
 ## 1.14.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 1.15.0
+
+* Re-export a number of platform library classes from `dart:collection`
+  in anticipation of moving the classes to this package for Dart 2.0.
+
 ## 1.14.1
 
 * Make `Equality` implementations accept `null` as argument to `hash`.
@@ -10,7 +15,7 @@
 
 ## 1.13.0
 
-* Add `EqualityBy`
+* Add `EqualityBy`.
 
 ## 1.12.0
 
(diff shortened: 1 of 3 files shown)


floitschG:
LGTM.

kevmoo:
You'll need to exclude dartfmt from stable, it seems.

Or only put it in `include`.

See https://github.com/dart-lang/test/blob/5750bd37e6642ef7854f7d9eee6c8581f0ca0993/.travis.yml#L60

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 1.15.0
+
+* Re-export a number of platform library classes from `dart:collection`
+  in anticipation of moving the classes to this package for Dart 2.0.
```
nex3: List the classes that are being re-exported, so users of those classes can search the changelog for a class name to see which version of the package first exposed that class.

lib/collection.dart
```diff
@@ -2,6 +2,20 @@
...
+    DoubleLinkedQueueEntry,
+    HasNextIterator,
+    LinkedList,
+    LinkedListEntry,
+    MapView, // TODO(lrn): Drop this class, it is equivalent to DelegatingMap.
+    SplayTreeMap,
+    SplayTreeSet,
+    UnmodifiableListView,
+    UnmodifiableMapBase,
+    UnmodifiableMapView;
```
nex3: The unmodifiable classes are widely used. I think we should consider leaving them in `dart:collection`.

pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: collection
-version: 1.14.4-dev
+version: 1.15.0
```
kevmoo: On 2nd thought, I'd hold off doing this. Copy-paste the classes and then deploy.

We really want an upper bound on the SDK constraint so folks don't get this package w/ the SDK that doesn't have these classes.

lib/collection.dart
lrhn: Being widly used isn't by itself enough to warrant a place in the platform libraries. We have `package:collection` (and friends) to hold generally useful code that doesn't *need* to be in the platform libraries.

pubspec.yaml
lrhn: There is an upper-bound on the SDK constraint of <2.0.0, and we will not remove the classes from Dart before 2.0.0. 

Is the problem that some 2.0.0-rcX will count as <2.0.0?

lib/collection.dart
nex3: These classes are [used occasionally](https://github.com/dart-lang/sdk/blob/cc5cb89e0b4ec788ad3b37d577f49ca7fa39b886/sdk/lib/io/platform_impl.dart#L110) in the SDK, but I'm also skeptical of the principle of stripping absolutely everything from the core libraries that's not strictly necessary for representing the transitive closure of the language spec and `dart:io`'s APIs or whatever.

The unmodifiable classes' APIs move in lock-step with the collection APIs, so they don't get any added flexibility from being in a package. They're not only widely-used, they're a fundamental part of exposing an API surface with strong encapsulation guarantees. I don't see the benefit of removing them from the core libraries, whereas requiring many users to depend on `collection` is a real cost.

matanlurey:
Do we still want to move here @lrhn @kevmoo?

lrhn:
Didn't happen for Dart 2. We'll reconsider for Dart 3.

PR #65, opened 2017-12-27
Add methods where needed for Dart 2.0 core lib changes.

### LINKED ISSUE #61: Add changelog entries for 1.14.3 + 1.14.2
Please?

@jmesserly @lrhn

srawlins:
Right now, all of the implementations throw (except in trivial cases, like the
EmptyUnmodifiable* classes). They can be updated, typically to call `super` or
reference a wrapped object, once the new methods have been released as part of a
Dart 2.0 release, and this pubspec.yaml updates to guarantee that those
implementations exist.

Also fixes #61

### DIFF
--- a/lib/src/canonicalized_map.dart
+++ b/lib/src/canonicalized_map.dart
@@ -69,6 +69,22 @@ class CanonicalizedMap<C, K, V> implements Map<K, V> {
     other.forEach((key, value) => this[key] = value);
   }
 
+  @override
+  // TODO: Dart 2.0 requires this method to be implemented.
+  // ignore: override_on_non_overriding_method
+  void addEntries(Iterable<Object> entries) {
+    // Change Iterable<Object> to Iterable<MapEntry<K, V>> when
+    // the MapEntry class has been added.
+    throw new UnimplementedError('addEntries');
+  }
+
+  @override
+  // TODO: Dart 2.0 requires this method to be implemented.
+  // ignore: override_on_non_overriding_method
+  Map<K2, V2> cast<K2, V2>() {
+    throw new UnimplementedError('cast');
+  }
+
   void clear() {
     _base.clear();
   }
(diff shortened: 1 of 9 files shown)


lib/src/canonicalized_map.dart
```diff
@@ -69,6 +69,22 @@ class CanonicalizedMap<C, K, V> implements Map<K, V> {
     other.forEach((key, value) => this[key] = value);
   }
 
+  @override
```
lrhn: This package do not use `@override`, so no need to add it here, or elsewhere.
That also voids the need for the "ignore:" below.

srawlins:
Thanks @lrhn ! `@override`s removed, and a few more impls added, after testing with SDK CL.

test/wrapper_test.dart
```diff
@@ -87,6 +87,9 @@ const TO_STRING_INVOCATION = const SyntheticInvocation(
 class IterableNSM extends NSM implements Iterable, Set, List, Queue {
   IterableNSM(action(Invocation i)) : super(action);
   toString() => super.noSuchMethod(TO_STRING_INVOCATION) as String;
+
+  Null cast<T>() => null;
+  Null retype<T>() => null;
```
lrhn: Make these methods abstract. The class uses `noSuchMethod` to handle (and record) everything in the `Iterable` interface, so it shouldn't need implementations. I can see that the return type needs fixing here since it must be both `List` and `Set` (and I should proabably rewrite the test to not use the same class for everything :)

lrhn:
Maybe update pubspec.yaml to make this a minor version increment, so we can publish it.

test/wrapper_test.dart
srawlins: Done.

srawlins:
Updated pubspec and changelog!

PR #67, opened 2018-01-30
Fix void usage for dart 2

MichaelRFairhurst:


MichaelRFairhurst:
Thanks for the quick review Kevin. Would it be useful or problematic if I synced this?

PR #69, opened 2018-02-15
Move flutter's CachingIterable to collection

natebosch:


natebosch:
cc @Hixie

Flutter has a dependency on this package:
https://github.com/flutter/flutter/blob/b3e4976ca20dc878bd8a484b877b66351bed86e4/packages/flutter/pubspec.yaml#L8

After this is publish you should be able to drop the copy within flutter.

natebosch:
Original: https://github.com/flutter/flutter/blob/c2e1bc0322605310511995f842db31fc5a83dc64/packages/flutter/lib/src/foundation/basic_types.dart

kevmoo:
While you're at it, add the 'build only master' bit to the travis file?

natebosch:
@kevmoo - #70

Hixie:
We're actually planning on going the other way, removing the dependency on collection entirely, eventually. Because we pin dependencies, every dependency we have is an additional constraint on our users.

natebosch:
In that case maybe it would be OK to duplicate it? We have a need in another package, I could duplicate it there but it seems like something that is generally useful...

Hixie:
Duplicating seems fine to me. We'll probably end up doing exactly that in reverse for the classes we use. :-)

test/caching_iterable_test.dart
```diff
@@ -0,0 +1,115 @@
...
+    yield index;
+  }
+}
+
+void main() {
+  setUp(() {
+    yieldCount = 0;
+  });
+
+  test('The Caching Iterable: length caches', () {
```
nex3: I'd get rid of these headers, since the test suite name makes it clear what's being tested. If you want to keep them, make them a group so you don't have to repeat them over and over.

test/caching_iterable_test.dart
```diff
@@ -0,0 +1,115 @@
...
+
+import 'package:test/test.dart';
+
+import 'package:collection/collection.dart';
+
+int yieldCount;
+
+Iterable<int> range(int start, int end) sync* {
+  assert(yieldCount == 0);
+  for (int index = start; index <= end; index += 1) {
```
nex3: `var index`

test/caching_iterable_test.dart
```diff
@@ -0,0 +1,115 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:test/test.dart';
+
+import 'package:collection/collection.dart';
+
+int yieldCount;
+
+Iterable<int> range(int start, int end) sync* {
```
nex3: Document this.

test/caching_iterable_test.dart
```diff
@@ -0,0 +1,115 @@
...
+  }
+}
+
+void main() {
+  setUp(() {
+    yieldCount = 0;
+  });
+
+  test('The Caching Iterable: length caches', () {
+    final Iterable<int> i = new CachingIterable<int>(range(1, 5).iterator);
```
nex3: `var iterable` or maybe `var integers` to match tests below.

The generic type in the constructor invocation shouldn't be necessary either.

test/caching_iterable_test.dart
```diff
@@ -0,0 +1,115 @@
...
+    expect(i.length, equals(5));
+    expect(yieldCount, equals(5));
+
+    expect(i.length, equals(5));
+    expect(yieldCount, equals(5));
+
+    expect(i.last, equals(5));
+    expect(yieldCount, equals(5));
+
+    expect(i, equals(<int>[1, 2, 3, 4, 5]));
```
nex3: This generic type is *definitely* not necessary.

test/caching_iterable_test.dart
```diff
@@ -0,0 +1,115 @@
...
+    expect(yieldCount, equals(5));
+  });
+
+  test('The Caching Iterable: laziness', () {
```
nex3: Test names should generally state the behavior that should exist. For example, I might call this "only gets as many values as necessary".

test/caching_iterable_test.dart
```diff
@@ -0,0 +1,115 @@
...
+    final Iterable<int> i = new CachingIterable<int>(range(1, 5).iterator);
+    expect(yieldCount, equals(0));
+
+    expect(i.first, equals(1));
+    expect(yieldCount, equals(1));
+
+    expect(i.firstWhere((int i) => i == 3), equals(3));
```
nex3: `(int i)` => `(value)`

test/caching_iterable_test.dart
```diff
@@ -0,0 +1,115 @@
...
+    expect(yieldCount, equals(3));
+
+    expect(i.last, equals(5));
+    expect(yieldCount, equals(5));
+  });
+
+  test('The Caching Iterable: where and map', () {
```
nex3: `where` and `map` should each have their own test. Same below.

test/caching_iterable_test.dart
```diff
@@ -0,0 +1,115 @@
...
+    expect(expanded1, equals(<int>[1, 1, 2, 2, 3, 3, 4, 4, 5, 5]));
+    expect(yieldCount, equals(5));
+
+    final Iterable<int> expanded2 = integers.expand((int i) => <int>[i, i]);
+
+    expect(yieldCount, equals(5));
+    expect(expanded2, equals(<int>[1, 1, 2, 2, 3, 3, 4, 4, 5, 5]));
+    expect(yieldCount, equals(5));
+  });
+}
```
nex3: Test `toList()` as well, and the explicit `Iterator` object.

natebosch:
If flutter isn't going to use this copy I know of only one use case which may end up being unnecessary after I measure performance. I'm just gonna write a smaller version where it's used.

PR #71, opened 2018-02-22
Remove usage of Maps.mapToString()

srawlins:
`Maps.mapToString()` is going to be removed by Dart 2.0. It will be moved to `MapBase.mapToString()`. To avoid bumping the minimum SDK version, we can just copy the implementation (~30 lines) to this package.


lrhn:
LGTM

nex3:
DBC: It would be good to document in a comment where this code comes from and why it's duplicated here.

PR #73, opened 2018-03-06
Fix suggested in issue https://github.com/dart-lang/sdk/issues/32415

a-siva:


CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
+## 1.14.6
+
+* Fix strong mode issue with ListEquality (the constructor of ListEquality
+  assumes that DefaultEquality will be inferred with generic type 'E' which
```
lrhn: They actually expected it to be filled in with `dynamic`. This is a Dart 1 behavior which is no longer valid in Dart 2.
So, maybe say it as:

Fix strong mode issue with `ListEquality`. The constructor of `ListEquality` needs a default value which is constant, assignable to `Equality<E>` for any `E` and which can compare elements of type `E`. This change makes `const DefaultEquality<Null>()` a solution to that constraint.

CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## 1.14.6
+
+* Fix strong mode issue with ListEquality. The constructor of ListEquality
+  needs a default value which is constant, assignable to Equality<E> for
+  any E and which can compare elements of type E. This change makes const
+  DefaultEquality<Null>() a solution to that constraint.
```
nex3: Nit: add backticks around type names and variables. This is especially important when writing generics, because otherwise the angle brackets will be interpreted as HTML tags and probably sanitized away.

```markdown
* Fix strong mode issue with `ListEquality`. The constructor of `ListEquality`
  needs a default value which is constant, assignable to `Equality<E>` for
  any `E` and which can compare elements of type `E`. This change makes
  `const DefaultEquality<Null>()` a solution to that constraint.
```

a-siva:
Thanks.

Issue #74, opened 2018-03-12
Implement Dart 2.0 collection operators


matanlurey:
See TODOs in https://github.com/dart-lang/collection/blob/master/lib/src/wrappers.dart.

Internal users are hitting this when attempting to use `.cast` and friends:

* [x] `Map.addEntries`
* [x] `Map.entries`
* [x] `Map.map`
* [x] `Map.cast`
* [x] `Map.retype`
* [x] `Map.update`
* [x] `Map.updateAll`
* [x] `Map.removeWhere`

* [x] `List.cast`
* [x] `List.+`

* [x] `Iterable.followedBy`
* [x] `Iterable.retype`
* [x] `Iterable.whereType`
* [x] `Iterable.singleWhere (orElse)`
* [x] `Iterable.indexWhere`
* [x] `Iterable.lastIndexWhere`

-----

NOTE: Adding tests for every one of these for everyone of the implementing classes... Not a fun experience. I'm going to be skipping `TypeSafeX` and `DelegatingX` when writing tests: https://github.com/dart-lang/collection/issues/81.

matanlurey:
I'll start tackling as it is blocking internal users.

Issue #81, opened 2018-03-12
Tests for "wrapper" implementations in #74


matanlurey:
See https://github.com/dart-lang/collection/issues/74.

I won't be writing tests for TypeSafeX or DelegatingX for Dart 2.0 collections, as it will be a huge time sink with little value. Perhaps this can be done in the future.

PR #84, opened 2018-03-12
Implement the rest of the TODOs.

matanlurey:
Closes https://github.com/dart-lang/collection/issues/74.

I am punting on the tests (https://github.com/dart-lang/collection/issues/81), since again, not clear of the value for most of these outside of code review, and will take a non-trivial amount of time.

/cc @natebosch

### DIFF
--- a/lib/src/typed_wrappers.dart
+++ b/lib/src/typed_wrappers.dart
@@ -40,10 +40,8 @@ abstract class _TypeSafeIterableBase<E> implements Iterable<E> {
       _base.fold(initialValue,
           (previousValue, element) => combine(previousValue, element as E));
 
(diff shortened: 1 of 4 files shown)


lib/src/typed_wrappers.dart
```diff
@@ -190,7 +178,6 @@ class TypeSafeList<E> extends TypeSafeIterable<E> implements DelegatingList<E> {
   int lastIndexOf(E element, [int start]) =>
       _listBase.lastIndexOf(element, start);
 
-  // TODO: Dart 2.0 requires this method to be implemented.
```
alorenzen: You removed the TODO, but this is still unimplemented...

matanlurey: Done, thanks!

Issue #80, opened 2018-03-12
Deprecate TypeSafeWrappers?


matanlurey:
... given that Dart 2 does this out of the box?

Here's what public APIs package offers today:

* [`DelegatingIterable.typed`][1]
* [`DelegatingList.typed`][2]
* [`DelegatingMap.typed`][3]
* [`DelegatingQueue.typed`][4]
* [`DelegatingSet.typed`][5]

[1]: https://www.dartdocs.org/documentation/collection/1.14.6/collection/DelegatingIterable/typed.html

[2]:
https://www.dartdocs.org/documentation/collection/1.14.6/collection/DelegatingList/typed.html

[3]:
https://www.dartdocs.org/documentation/collection/1.14.6/collection/DelegatingMap/typed.html

[4]:
https://www.dartdocs.org/documentation/collection/1.14.6/collection/DelegatingQueue/typed.html

[5]:
https://www.dartdocs.org/documentation/collection/1.14.6/collection/DelegatingSet/typed.html

Here's an example of the implementation of those static methods:

https://github.com/dart-lang/collection/blob/fac9e50be7e61819e66abfd6381d55135b8751d3/lib/src/wrappers.dart#L148-L149

I imagine to avoid churn, we could do the following:

* Add a package-internal field called `isStrongMode`:

```dart
final bool isStrongMode = (() {
  bool typeParamIsString<T>(T item) => T == String;
  return typeParamIsString('');
})();
```

* Change the `.typed` methods to be:

```dart
static List<E> typed<E>(List base) {
  if (isStrongMode) {
    return base.cast<E>();
  }
  return base is List<E> ? base : new TypeSafeList<E>(base);
}
```

* When Dart 2 is enabled by default everywhere (Dart2JS/VM/Flutter), remove the `TypeSafeX` wrapper classes, as they will never be used. This should not be a breaking change, strictly. 

* At the same time of above, deprecate the `.typed` methods, because they no longer have value and are misleading, doing `is` and `.cast` is fine for all users at this point.

Thoughts? @lrhn @srawlins @kevmoo

nex3:
As long as we're only supporting Dart 2 (which appears to be the case from the other changes), we can do this much more simply by just using `cast()` everywhere. This is already a no-op in Dart 1 mode.

matanlurey:
Definitely, I just am trying to be nice to users - in my own tests it looks like the type of errors thrown by Dart2 versus thrown currently by the wrappers are not exactly the same.

[...]

PR #86, opened 2018-03-13
Deleted typed wrappers, use Dart 2 instead.

matanlurey:
Closes https://github.com/dart-lang/collection/issues/80.


CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 1.14.8
+
+* Deprecated `Delegating{Name}.typed` static methods in favor of the new Dart 2
+  `cast` methods. For example, `DelegatingList.typed<String>(list)` can now just
```
natebosch: [nit] remove 'just'

matanlurey:
> How much confidence do we have that this isn't breaking, even for users of DDC?

Not 100%, but given the amount of usage, I'd rather wait for bug reports and partially revert if we need to. I'll keep my eye out.

Issue #87, opened 2018-03-16
QueueList's cast and retype do not adhere to the List methods' documentation


srawlins:
I'd like to say that QueueList's `cast` and `retype` do not adhere to the _contract_ laid out by `List.cast` and `List.retype` but that may be up for debate. In any case, `List.cast`'s documentation says:

> Returns a view of this list as a list of `R` instances, if necessary.

`List.retype`'s documentation says, similarly:

> Returns a view of this list as a list of `R` instances.

(Ignoring for now what "if necessary" must mean...) I believe it is intended that the return value be *the same list* (just a different view of it), **not** a new list. For example, let's play with `List`:

```dart
void main() {
  var a = new List<num>()..addAll([1,2,3]);
  var b = a.retype<int>();
  b.add(4); // add 4 to the view.
  a.add(5); // add 5 to the original.
  b.remove(3); // remove 3 from the view.
  print('a: $a'); //=> [1, 2, 4, 5]
  print('b: $b'); //=> [1, 2, 4, 5]
}
```

They are the same list. Now let's do the same for QueueList:

```dart
import 'package:collection/collection.dart';

void main() {
  var c = new QueueList<num>()..addAll([1,2,3]);
  var d = c.retype<int>();
  d.add(4);
  print('c: $c'); //=> {1, 2, 3} // Not the same list!
  print('d: $d'); //=> {1, 2, 3, null, null, null, null, null, 4} // Duck?
  c.add(5);
  print('c: $c'); //=> {1, 2, 3, 5}
  print('d: $d'); //=> {1, 2, 3, null, null, null, null, null, 4}
}
```

I have not checked the other `cast` and `retype` implementations yet, but when I do, I presume I'll just add on to this bug, rather than filing a bug for each class.

matanlurey:
> I believe it is intended that the return value be *the same list* (just a different view of it), **not** a new list.

I don't think so (@leafpetersen @lrhn), even though I wish that was the case - I think `.cast` _tries_ to re-use the same list, and basically invokes `.retype` if not possible, making this OK. 

I.e. this is just a convenience for calling `List.castFrom`:

https://github.com/dart-lang/sdk/blob/f1ebe2bd5cfcb6b522e5b4fd406cdabb1a2d2091/sdk/lib/collection/list.dart#L332-L335

Let's look at the implementation of `DoubleLinkedQueue.cast`:

[...]

PR #88, opened 2018-03-19
Fix cast and retype implementations.

matanlurey:
Closes https://github.com/dart-lang/collection/issues/87.

I had to be a little creative for `QueueList`. The others seem OK enough.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,16 @@
+## 1.15.0
+
+* Added `QueueList.castFrom<S, T>` (similar to `List.castFrom`, `Queue.castFrom`).
+
+* Added `MapKeySet.castFrom<S, T>` (similar to `Set.castFrom`).
+
+* Added `MapValueSet.castFrom<K, V, K2, V2>` (similar to `Map.castFrom`).
+
+* Fixed bugs where `QueueList`, `MapKeySet`, and `MapValueSet` did not adhere to
+  the contract laid out by `List.cast`, `Set.cast` and `Map.cast` respectively.
+  The returned instances of these methods now correctly forward to the existing
+  instance instead of always creating a new copy.
+
 ## 1.14.8
 
 * Deprecated `Delegating{Name}.typed` static methods in favor of the new Dart 2
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,6 @@
 
 * Added `QueueList.castFrom<S, T>` (similar to `List.castFrom`, `Queue.castFrom`).
 
-* Added `MapKeySet.castFrom<S, T>` (similar to `Set.castFrom`).
-
-* Added `MapValueSet.castFrom<K, V, K2, V2>` (similar to `Map.castFrom`).
-
 * Fixed bugs where `QueueList`, `MapKeySet`, and `MapValueSet` did not adhere to
   the contract laid out by `List.cast`, `Set.cast` and `Map.cast` respectively.
   The returned instances of these methods now correctly forward to the existing
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,5 @@
 ## 1.15.0
 
-* Added `QueueList.castFrom<S, T>` (similar to `List.castFrom`, `Queue.castFrom`).
-
 * Fixed bugs where `QueueList`, `MapKeySet`, and `MapValueSet` did not adhere to
   the contract laid out by `List.cast`, `Set.cast` and `Map.cast` respectively.
   The returned instances of these methods now correctly forward to the existing
--- a/lib/src/queue_list.dart
+++ b/lib/src/queue_list.dart
@@ -21,7 +21,7 @@ class QueueList<E> extends Object with ListMixin<E> implements Queue<E> {
   /// If all accessed elements of [source] are actually instances of [T] and if
   /// all elements stored in the returned  are actually instance of [S],
   /// then the returned instance can be used as a `QueueList<T>`.
-  static QueueList<T> castFrom<S, T>(QueueList<S> source) {
+  static QueueList<T> _castFrom<S, T>(QueueList<S> source) {
     return new _CastQueueList<S, T>(source);
   }
 
@@ -106,7 +106,7 @@ class QueueList<E> extends Object with ListMixin<E> implements Queue<E> {
     return retype<T>();
   }
 
-  QueueList<T> retype<T>() => QueueList.castFrom<E, T>(this);
+  QueueList<T> retype<T>() => QueueList._castFrom<E, T>(this);
 
   String toString() => IterableBase.iterableToFullString(this, "{", "}");
 
--- a/lib/src/queue_list.dart
+++ b/lib/src/queue_list.dart
@@ -99,10 +99,11 @@ class QueueList<E> extends Object with ListMixin<E> implements Queue<E> {
   }
 
   QueueList<T> cast<T>() {
-    if (this is QueueList<T>) {
-      return this as QueueList<T>;
+    QueueList<Object> self = this;
+    if (self is QueueList<T>) {
+      return self;
     }
-    return QueueList.castFrom<E, T>(this);
+    return retype<T>();
   }
 
   QueueList<T> retype<T>() => QueueList.castFrom<E, T>(this);
--- a/lib/src/queue_list.dart
+++ b/lib/src/queue_list.dart
@@ -10,6 +10,21 @@ import 'dart:collection';
 // that are redundant with ListMixin. Remove or simplify it when issue 21330 is
 // fixed.
 class QueueList<E> extends Object with ListMixin<E> implements Queue<E> {
+  /// Adapts [source] to be a `QueueList<T>`.
+  ///
+  /// Any time the class would produce an element that is not a [T], the element
+  /// access will throw.
+  ///
+  /// Any time a [T] value is attempted stored into the adapted class, the store
+  /// will throw unless the value is also an instance of [S].
+  ///
+  /// If all accessed elements of [source] are actually instances of [T] and if
+  /// all elements stored in the returned  are actually instance of [S],
+  /// then the returned instance can be used as a `QueueList<T>`.
+  static QueueList<T> castFrom<S, T>(QueueList<S> source) {
+    return new _CastQueueList<S, T>(source);
+  }
+
   static const int _INITIAL_CAPACITY = 8;
   List<E> _table;
   int _head;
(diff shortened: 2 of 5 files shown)


lib/src/wrappers.dart
```diff
@@ -469,13 +469,33 @@ class DelegatingMap<K, V> implements Map<K, V> {
...
+  /// access will throw.
+  ///
+  /// Any time a [T] value is attempted stored into the adapted set, the store
+  /// will throw unless the value is also an instance of [S].
+  ///
+  /// If all accessed elements of [source] are actually instances of [T] and if
+  /// all elements stored in the returned  are actually instance of [S],
+  /// then the returned set can be used as a `MapKeySet<T>`.
+  static MapKeySet<T> castFrom<S, T>(MapKeySet<S> source) {
+    return new MapKeySet(source._baseMap.cast<T, dynamic>());
```
lrhn: Too creative here. A `MapKeySet` is just as `Set`, so use `Set.castFrom` and don't introduce a new  `castFrom` for this implementation of `Set`.

lib/src/wrappers.dart
```diff
@@ -547,6 +567,24 @@ class MapKeySet<E> extends _DelegatingIterableBase<E>
...
+  /// Any time the set would produce an element that is not a [T], the element
+  /// access will throw.
+  ///
+  /// Any time a [T] value is attempted stored into the adapted set, the store
+  /// will throw unless the value is also an instance of [S].
+  ///
+  /// If all accessed elements of [source] are actually instances of [T] and if
+  /// all elements stored in the returned  are actually instance of [S],
+  /// then the returned set can be used as a `MapKeySet<T>`.
+  static MapValueSet<K2, V2> castFrom<K, V, K2, V2>(MapValueSet<K, V> source) {
```
lrhn: Again, this class is just an implementation of `Set`, so just use `Set.castFrom`.

lib/src/queue_list.dart
```diff
@@ -84,10 +102,10 @@ class QueueList<E> extends Object with ListMixin<E> implements Queue<E> {
     if (this is QueueList<T>) {
```
lrhn: While you are here, please rewrite this as:
```
QueueList<Object> self = this;
return self is QueueList<T> ? self : this.retype<T>();
```
It generates better code for dart2js to not have an explicit cast. By up-casting first. we make sure that a successful test promotes the variable.

Ditto for the `cast` methods below.

CHANGELOG.md
```diff
@@ -1,3 +1,16 @@
+## 1.15.0
+
+* Added `QueueList.castFrom<S, T>` (similar to `List.castFrom`, `Queue.castFrom`).
+
+* Added `MapKeySet.castFrom<S, T>` (similar to `Set.castFrom`).
```
lrhn: No need for the last two `castFrom`'s.

lib/src/wrappers.dart
```diff
@@ -469,13 +469,33 @@ class DelegatingMap<K, V> implements Map<K, V> {
...
+  }
+
   final Map<E, dynamic> _baseMap;
 
   MapKeySet(Map<E, dynamic> base) : _baseMap = base;
 
   Iterable<E> get _base => _baseMap.keys;
 
-  Set<T> cast<T>() => _baseMap.keys.toSet().cast<T>();
+  MapKeySet<T> cast<T>() {
```
lrhn: Just return `Set<T>` and use `Set.castFrom`.

lib/src/wrappers.dart
```diff
@@ -513,7 +533,7 @@ class MapKeySet<E> extends _DelegatingIterableBase<E>
   E lookup(Object element) =>
       throw new UnsupportedError("MapKeySet doesn't support lookup().");
 
-  Set<T> retype<T>() => _baseMap.keys.toSet().retype<T>();
+  Set<T> retype<T>() => MapKeySet.castFrom<E, T>(this);
```
lrhn: `MapKeySet` -> `Set`

lib/src/queue_list.dart
```diff
@@ -230,3 +248,17 @@ class QueueList<E> extends Object with ListMixin<E> implements Queue<E> {
     _head = 0;
   }
 }
+
+class _CastQueueList<S, T> extends QueueList<T> {
```
lrhn: This is creative. Tests, or it doesn't work. :)

srawlins: I was hoping to `extend ThingImAddingRetypeTo` as well 😄

lib/src/wrappers.dart
matanlurey: Done.

lib/src/wrappers.dart
matanlurey: Done.

lib/src/queue_list.dart
matanlurey: Done.

CHANGELOG.md
matanlurey: Done.

lib/src/wrappers.dart
matanlurey: Done.

lib/src/wrappers.dart
matanlurey: Done.

lib/src/queue_list.dart
matanlurey: Done. Take a look - tests are basic.

test/queue_list_test.dart
```diff
@@ -258,6 +258,25 @@ void main() {
...
+    expect(stringQueue, ['a', 'b', 'c', 'd']);
+  });
+
+  test("cast throws on mutation when the type is not valid", () {
+    QueueList<Object> stringQueue = new QueueList<String>();
+    var numQueue = stringQueue.cast<num>();
+    expect(
+      () => numQueue.add(1),
+      throwsCastError,
+      skip: isDart2 ? false : 'In checked mode (VM), CastError is not thrown',
```
lrhn: The real reason is that `cast` only works in Dart 2 because it depends on the type parameter of a generic method. It's not VM specific in any way.

(In checked mode, the `add(1)` *would* throw because you are adding an `int` to a `QueueList<String>`, but it's more likely to be a `TypeError` than a `CastError`).

lib/src/queue_list.dart
```diff
@@ -99,10 +99,11 @@ class QueueList<E> extends Object with ListMixin<E> implements Queue<E> {
...
-      return this as QueueList<T>;
+    QueueList<Object> self = this;
+    if (self is QueueList<T>) {
+      return self;
     }
-    return QueueList.castFrom<E, T>(this);
+    return retype<T>();
   }
 
   QueueList<T> retype<T>() => QueueList.castFrom<E, T>(this);
```
lrhn: TBH, after some more consideration, I'd remove `QueueList.castFrom` again.
It's a very specific implementation class, not a general interface that others might implement, so you don't need to make the wrapping available to to others.
The reason for `Iterable.castFrom` is that it can be used to implement `Iterable.cast` for other classes that *implement* `Iterable`, but nobody should implement `QueueList`, they should just implement `Queue` and `List` if they want a class that is both.

So, consider making `castFrom` private or just inline it here.

test/queue_list_test.dart
```diff
@@ -258,6 +258,25 @@ void main() {
     expect(queue.cast<Pattern>(), same(queue));
   }, skip: isDart2 ? false : 'Requires a Dart2 runtime');
 
+  test("cast does not throw on mutation when the type is valid", () {
+    var patternQueue = new QueueList<Pattern>()..addAll(['a', 'b']);
+    QueueList<Object> untypedQueue = patternQueue;
```
lrhn: Why this line? Changing the static type has no effect on `cast`, you can just write `patternQueue.cast<String>()` below.

test/queue_list_test.dart
```diff
@@ -258,6 +258,25 @@ void main() {
...
+    var stringQueue = untypedQueue.cast<String>();
+    stringQueue.addAll(['c', 'd']);
```
lrhn: If you forgot to do the `cast`, the test would still succeed (either because it infers `<Pattern>['c', 'd']`, or because `List<String>` is-a `List<Pattern>`).

At least add `expect(stringQueue is List<String>, true);`, which shows that the `cast` actually did *something* (and skip in non-dart2 mode because there `cast` actually doesn't do anything).

test/queue_list_test.dart
```diff
@@ -258,6 +258,25 @@ void main() {
...
+
+    expect(stringQueue, ['a', 'b', 'c', 'd']);
```
lrhn: Check that `patternQueue` was also modified.

test/queue_list_test.dart
matanlurey: Thanks. Done!

lib/src/queue_list.dart
matanlurey: Made it private. Thanks!

test/queue_list_test.dart
matanlurey: Done.

test/queue_list_test.dart
matanlurey: Done.

test/queue_list_test.dart
matanlurey: Done.

test/queue_list_test.dart
matanlurey: Looks like the type thrown is `CastErrorImplementation`, which is not a `TypeError`.

test/queue_list_test.dart
```diff
@@ -260,20 +260,36 @@ void main() {
 
   test("cast does not throw on mutation when the type is valid", () {
     var patternQueue = new QueueList<Pattern>()..addAll(['a', 'b']);
-    QueueList<Object> untypedQueue = patternQueue;
-    var stringQueue = untypedQueue.cast<String>();
+    var stringQueue = patternQueue.cast<String>();
     stringQueue.addAll(['c', 'd']);
+    expect(
+      stringQueue,
+      const isInstanceOf<QueueList<Pattern>>(),
```
lrhn: Check for `QueueList<String>` here. That's the type you are casting to, and a type that `patternQueue` *doesn't* have, so it checks that something did happen.
(You could throw in a check that `stringQueue` and `patternQueue` are not *identical*, but it's not important after checking that `stringQueue` has a type that `patternQueue` doesn't).

This is the `expect` that needs the `skip` parameter, because in Dart 1 mode, the type check won't be true since `cast` returns `patternQueue` unmodified. The two tests below will still succeed when `stringQueue` and `patternQueue` are the same object.

matanlurey: Done. Thanks @lrhn!

Issue #91, opened 2018-05-01
Dart 2 runtime failures in tests


keertip:
running tests in dart 2 mode has failures

Expected: throws UnsupportedError
  Actual: <Closure: () => Null>
   Which: threw ?:<type '_GeneratorIterable<dynamic>' is not a subtype of type 'Iterable<int>' of 'iterable' where
                  _GeneratorIterable is from dart:core
                  Iterable is from dart:core
                  int is from dart:core
                >
          stack dart:collection                                                                               _ListBase&Object&ListMixin.setRange
                ../../collection/test/unmodifiable_collection_test.dart 338:13  testNoWriteList.<fn>

type '_CompactLinkedHashSet<dynamic>' is not a subtype of type 'Set<int>' of 'component' where...
type '_CompactLinkedHashSet<dynamic>' is not a subtype of type 'Set<int>' of 'component' where
  _CompactLinkedHashSet is from dart:collection
  Set is from dart:core
  int is from dart:core
package:collection/src/union_set_controller.dart                                          UnionSetController.add
../../collection/test/union_set_controller_test.dart 20:16  main.<fn>

type '_CompactLinkedHashSet<dynamic>' is not a subtype of type 'Set<int>' of 'other' where...
type '_CompactLinkedHashSet<dynamic>' is not a subtype of type 'Set<int>' of 'other' where
  _CompactLinkedHashSet is from dart:collection
  Set is from dart:core
  int is from dart:core
package:collection/src/wrappers.dart                                                          DelegatingSet.union
../..dart/collection/test/unmodifiable_collection_test.dart 476:20  testReadSet.<fn>

and more....

PR #93, opened 2018-05-02
Fix 2 tests

lrhn:
Fix two tests to be Dart 2 type-safe.

Clean up test/wrapper_test.dart signficantly.
Add enough type annotations to test/unmodifiable_collection_test.dart to get the necessary inference.
Update dependency on SDK to allow using the new `Invocation` constructor.

Another partial fix of #91.

### DIFF
--- a/test/unmodifiable_collection_test.dart
+++ b/test/unmodifiable_collection_test.dart
@@ -311,8 +311,8 @@ void testReadList(List original, List wrapped, String name) {
   });
 }
 
-void testNoWriteList(List original, List wrapped, String name) {
-  List copy = new List.from(original);
+void testNoWriteList(List<int> original, List<int> wrapped, String name) {
+  var copy = new List<int>.from(original);
 
   testThrows(name, thunk) {
     test(name, () {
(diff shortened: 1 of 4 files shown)


test/unmodifiable_collection_test.dart
```diff
@@ -311,8 +311,8 @@ void testReadList(List original, List wrapped, String name) {
   });
 }
 
-void testNoWriteList(List original, List wrapped, String name) {
-  List copy = new List.from(original);
+void testNoWriteList(List<int> original, List<int> wrapped, String name) {
+  var copy = new List<int>.from(original);
```
nex3: Can you use `.of()` here and below to avoid the explicit annotation?

test/wrapper_test.dart
```diff
@@ -25,261 +25,245 @@ void testInvocations(Invocation i1, Invocation i2) {
...
-  final List positionalArguments;
-  final Map<Symbol, dynamic> namedArguments;
-  final int _type;
-  const SyntheticInvocation(this.memberName, this.positionalArguments,
-      this.namedArguments, this._type);
-
-  List<Type> get typeArguments => const <Type>[];
-
-  bool get isMethod => _type == METHOD;
+// Parameterization of noSuchMethod. Calles [_action] on every
```
nex3: "Calles" -> "Calls"

test/unmodifiable_collection_test.dart
lrhn: Seems so. Good catch.

test/wrapper_test.dart
lrhn: Whoops. Thanks.

PR #108, opened 2019-07-10
Fix lints

kevmoo:


kevmoo:
See https://travis-ci.org/dart-lang/collection/jobs/556612316

Issue #109, opened 2019-08-14
CombinedMapView does not maintain keys uniqueness


renatoathaydes:
The following test fails:

```dart
final map = CombinedMapView([{'1': 1, '2': 2, '3': 3}, {'2': 22, '4': 44}]);
expect(map.length, equals(4));
expect(map.keys.toList(), equals(['1', '2', '3', '4']));
```

The first expectation fails because `length` returns `5`.
The second, because `keys` returns `['1', '2', '3', '2', '4']`.

This seems to violate `Map`'s interface, which says:

```
There is a finite number of keys in the map, and each key has exactly one value associated with it.
```

This was a problem for me because I needed something that looks just like a normal `Map` but based on two other Maps, and which lets a map coming first "override" values on the map coming later in the list of maps. I believe that's the main purpose of combining maps (possibly unmodifiable), but due to this issue, it can't be used like that.

renatoathaydes:
My workaround was creating my own type on top of it:


```dart
class SnapshotMap<K, V> extends CombinedMapView<K, V> {
  Iterable<K> Function() _lazyKeys;

  SnapshotMap(Map<K, V> topMap, Map<K, V> bottomMap)
      : super([topMap, bottomMap]) {
    _lazyKeys = lazy(() => super.keys.toSet());
  }

  UnmodifiableMapView<K, V> get view => UnmodifiableMapView(this);

  @override
  Iterable<K> get keys => _lazyKeys();

  @override
  int get length => keys.length;
}
```

jakemac53:
Ya, I am going to see if I can maintain the lazy behavior of the current implementation though if possible. Otherwise this would likely break some projects.

renatoathaydes:
No the prettiest thing, but my `lazy` function above:

```dart
T Function() lazy<T>(T Function() f) {
  T instance;
  return () {
    if (instance == null) {
      instance = f();
      if (instance == null) {
        throw StateError('lazy function must not return null');
      }
    }
    return instance;
  };
}
```

jakemac53:
The problem there is the `toSet` call ends up eagerly iterating all the keys of all the maps as soon as you ask for the first key - I have a simple implementation coming though.

jakemac53:
Note that in adding the test for this I also noticed `forEach` and `values` were broken, fun!

jakemac53:
(these all get fixed by fixing `keys` though)

renatoathaydes:
[...]

PR #110, opened 2019-08-14
CombinedMapView.keys fix

jakemac53:
Fixes https://github.com/dart-lang/collection/issues/109

Adds a custom iterable/iterator that can filter out duplicates and use that for the `CombineMapView.keys` getter. 

Update tests to contain duplicates in maps, and ensure the keys/values from the earlier maps are the ones that are returned.

Updates the changelog and docs to no longer claim `O(maps)` for the length getter. This now requires iteration of all items and is `O(total map entries)`.

Prepare to publish as 1.14.12

### DIFF
--- a/lib/src/combined_wrappers/combined_map.dart
+++ b/lib/src/combined_wrappers/combined_map.dart
@@ -13,10 +13,12 @@ import 'combined_iterable.dart';
 /// accessing individual map instances. In the occasion where a key occurs in
 /// multiple maps the first value is returned.
 ///
-/// The resulting map has an index operator (`[]`) and `length` property that
-/// are both `O(maps)`, rather than `O(1)`, and the map is unmodifiable - but
-/// underlying changes to these maps are still accessible from the resulting
-/// map.
+/// The resulting map has an index operator (`[]`) that is `O(maps)`, rather
+/// than `O(1)`, and the map is unmodifiable, but underlying changes to these
+/// maps are still accessible from the resulting map.
+///
+/// The `length` is `O(maps.reduce((value, map) => value + map.length))` since
+/// it has to remove duplicate entries.
 class CombinedMapView<K, V> extends UnmodifiableMapBase<K, V> {
   final Iterable<Map<K, V>> _maps;
 
@@ -39,8 +41,10 @@ class CombinedMapView<K, V> extends UnmodifiableMapBase<K, V> {
 
   /// The keys of [this].
   ///
-  /// The returned iterable has efficient `length` and `contains` operations,
-  /// based on [length] and [containsKey] of the individual maps.
+  /// The returned iterable has efficient `contains` operations, based on
+  /// [containsKey] of the individual maps.
+  ///
+  /// The `length` must do deduplication and thus is not optimized.
   ///
   /// The order of iteration is defined by the individual `Map` implementations,
   /// but must be consistent between changes to the maps.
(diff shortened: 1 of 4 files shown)


jakemac53:
Hmm actually I realized this doesn't work if you iterate `keys` multiple times, it will always be empty after the first iteration.

jakemac53:
Ok, pushed a new implementation using a custom Iterable/Iterator view - still doesn't seem to terrible.

We could also exposing this new deduplicating iterable if we thought it was generally useful?

lib/src/combined_wrappers/combined_map.dart
```diff
@@ -39,14 +41,56 @@ class CombinedMapView<K, V> extends UnmodifiableMapBase<K, V> {
...
   ///
   /// The order of iteration is defined by the individual `Map` implementations,
   /// but must be consistent between changes to the maps.
   ///
   /// Unlike most [Map] implementations, modifying an individual map while
   /// iterating the keys will _sometimes_ throw. This behavior may change in
   /// the future.
-  Iterable<K> get keys => CombinedIterableView<K>(_maps.map((m) => m.keys));
+  Iterable<K> get keys => _LazyDeduplicatingIterableView(
+      CombinedIterableView(_maps.map((m) => m.keys)));
```
natebosch: Is the `CombinedIterableView` buying us much over `expand`? Does it get us more than an `isEmpty` optimization?

I don't think we need `_LazyDeduplicatingIterableView` if we use `sync*` since that will already re-invoke the function on each retrieval of the iterator. Not sure how much we want to try to micro-optimize.

```dart
Iterable<K> get keys sync* {
  final seen = <K>{};
  yield* _maps.expand((m) => m.keys).where(seen.add);
}
```

natebosch: Hmm, the `contains` optimization is a little more compelling...

lib/src/combined_wrappers/combined_map.dart
```diff
@@ -39,14 +41,56 @@ class CombinedMapView<K, V> extends UnmodifiableMapBase<K, V> {
 
   /// The keys of [this].
   ///
-  /// The returned iterable has efficient `length` and `contains` operations,
-  /// based on [length] and [containsKey] of the individual maps.
+  /// The returned iterable has efficient `contains` operations, based on
+  /// [containsKey] of the individual maps.
```
natebosch: This comment isn't quite accurate. We optimize based on `keys.contains` of the individual maps, we _expect_ those are in turn optimized based on `containsKey`, but we can't know for sure. Is it worth making a distinction?

jakemac53: I had a similar thought.... but I didn't feel like it was worth changing the existing comment?

lib/src/combined_wrappers/combined_map.dart
jakemac53: I think the contains optimization is worth it for something in a "core" lib like this.

lib/src/combined_wrappers/combined_map.dart
jakemac53: I went ahead and updated it

jakemac53:
PTAL, did some minor refactors

jakemac53:
Given that this is a potentially breaking bug fix I am going to test things internally and make sure the breakage isn't too bad before landing this.

jakemac53:
There is at least one breakage internally, I am still planning on releasing this change but it will be delayed just a little bit while I investigate the right course of action.

jakemac53:
Nevermind, breakages were not actually related even though they seemed like it, going to go ahead and merge/publish.

PR #117, opened 2020-02-02
Refactor UnionSet deduplicating iterable

natebosch:
`Set.add` returns a boolean so it is unnecessary to use both
`Set.contains` and then `Set.add`. Once the `.where` call is shortened
to a tearoff the deduplication can be inlined with `_iterable`.

### DIFF
--- a/lib/src/union_set.dart
+++ b/lib/src/union_set.dart
@@ -52,22 +52,13 @@ class UnionSet<E> extends SetBase<E> with UnmodifiableSetMixin<E> {
   @override
   Iterator<E> get iterator => _iterable.iterator;
 
-  /// Returns an iterable over the contents of all the sets in [this].
-  Iterable<E> get _iterable =>
-      _disjoint ? _sets.expand((set) => set) : _dedupIterable;
-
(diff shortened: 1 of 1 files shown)


kevmoo:
@natebosch – how'd you find THIS?!?!

natebosch:
Saw some usage, wasn't familiar with the class and I was curious how much it was accomplishing. Jumped in to look at the implementation and the contains then add pattern stood out to me.

lib/src/union_set.dart
```diff
@@ -52,22 +52,13 @@ class UnionSet<E> extends SetBase<E> with UnmodifiableSetMixin<E> {
...
-    return _sets.expand((set) => set).where((element) {
-      if (seen.contains(element)) return false;
-      seen.add(element);
-      return true;
-    });
+  /// If this is not a [_disjoint] union an extra set is used to deduplicate
+  /// values.
+  Iterable<E> get _iterable {
+    var allElements = _sets.expand((set) => set);
+    return _disjoint ? allElements : allElements.where(<E>{}.add);
```
jakemac53: While we are in here we might as well also make this a `HashSet`

PR #120, opened 2020-02-25
Implement BitArray.

AAAbramenko:
Implemented as non-growable list of bool.


tatumizer:
"cardinality" is a confusing name IMO. The number of ones in a bitmask is
conventionally called "population count"

lib/src/bitarray.dart
```diff
@@ -0,0 +1,317 @@
...
+    }
+    return result;
+  }
+
+  /// Replaces integer values in [_data] by applying [func] to each word.
+  void _transformDataWords(int Function(int word, int wordIndex) func) {
+    for (var wordIndex = 0; wordIndex < _data.length; ++wordIndex) {
+      _data[wordIndex] = func(_data[wordIndex], wordIndex);
+    }
+    _clearTailBits();
```
lrhn: This function updates by doing a function call per word. Most operations are so simple that inlining the loop and operation will still be almost a one-liner, and it will probably be quite a lot more efficient (unless the compiler manages to inline this function everywhere).

This is general reusable library code, it should be very efficient as a base-line because users who depend on it cannot change its performance.

lib/src/bitarray.dart
```diff
@@ -0,0 +1,317 @@
...
+/// Implemented as fixed-length [List] of [bool].
+/// Throws error when calling any operation with length modifying.
+///
+/// Support several ways to perform logical operations (NOT, OR, AND, XOR):
+/// * instance methods (mutate [this]): [BitArray.not], [BitArray.or], [BitArray.and], [BitArray.xor];
+/// * producing new instance of [BitArray]:
+///     * logical operators: [BitArray.~], [BitArray.|], [BitArray.&], [BitArray.^].
+///     * logical gates (as static methods): [BitArray.notGate], [BitArray.orGate], [BitArray.andGate], [BitArray.xorGate].
+/// All arguments for not-unary logical operations must have the same [length] or [ArgumentError] will be thrown.
+class BitArray with ListMixin<bool>, NonGrowableListMixin<bool> {
```
lrhn: If keeping it as a list, call it "BitList' or `BoolList`. We don't use "Array" anywhere else.

This is both an implementation and an interface (it introduces new methods on top of `List`).
As an implementation class, it's fine to refer to the implementation strategy (`Bit`) rather than the API (`bool`). A reusable interface should not be mentioning implementation details.
So, let's assume that this is not a reusable interface, there are not expected to be other implementations than this one. Then I'd call it `BitList`.

lib/src/bitarray.dart
```diff
@@ -0,0 +1,317 @@
...
+    }
+  }
+
+  /// Sets the bit at the [index] to the complement of its current value.
+  void flip(int index) {
+    RangeError.checkValidIndex(index, this);
+    _data[index >> 5] ^= 1 << (index & 31);
+  }
+
+  /// Sets each bit at the range from [start] inclusive to [end] exclusive to the complement of its current value.
```
lrhn: Shorter first sentence, then elaborate below. Maybe:

/// Toggles each bit in the range from [start] to [end].
///
/// Sets each bit in the range from [start] (inclusive) to [end] (exclusive) to the complement
/// of its current value.
///
/// The [start] and [end] must define a valid range of this list (0 &le; [start] &le; [end] &le; [length]).
/// If [end] is omitted, it defaults to [length].

lib/src/bitarray.dart
```diff
@@ -0,0 +1,317 @@
...
+  void flipRange(int start, [int end]) {
```
lrhn: I'd use `toggle` instead of `flip` for changing between two values.
Is "flip" a term of art in bit operations?

lib/src/bitarray.dart
```diff
@@ -0,0 +1,317 @@
...
+    end ??= _length;
+    RangeError.checkValidRange(start, end, _length);
+
+    var startWordIndex = start >> 5;
+    var endWordIndex = end >> 5;
+
+    if (startWordIndex == endWordIndex) {
+      for (var pos = start; pos < end; ++pos) {
+        flip(pos);
+      }
```
lrhn: Or maybe:
```dart
var bitStart = start & 31;
var bitEnd = end & 31;
_data[startWordIndex] ^= ((1 << (bitEnd - bitStart)) - 1) << bitStart;
```

lib/src/bitarray.dart
```diff
@@ -0,0 +1,317 @@
...
+      _data[endWordIndex] ^= 0xFFFFFFFF >> (32 - (end & 31));
+    }
+  }
+
+  @override
+  Iterator<bool> get iterator => BitArrayIterator(this);
+}
+
+/// Iterator for [BitArray].
+class BitArrayIterator implements Iterator<bool> {
```
lrhn: You can make this class private. Nobody needs to access it except the BitArray class.

lib/src/bitarray.dart
```diff
@@ -0,0 +1,317 @@
...
+  int _pos = 0;
+  final BitArray _bitArray;
+
+  BitArrayIterator(this._bitArray);
+
+  @override
+  bool get current => _current;
+
+  @override
+  bool moveNext() {
```
lrhn: If you want a custom iterator anyway, consider preparing for Null Safety by specifying that the iterator returns `false` before calling `moveNext` or after `moveNext` has returned false, instead of `null`.
(This should be specified on the `iterator` getter of `BitArray` if this class is private).

Then initialize `_current` to `false` and set it to `false` instead of `null` below.

lib/src/bitarray.dart
```diff
@@ -0,0 +1,317 @@
...
+    RangeError.checkValidIndex(index, this);
+    return (_data[index >> 5] & (1 << (index & 31))) != 0;
+  }
+
+  @override
+  int get length => _length;
+
+  @override
+  void operator []=(int index, bool val) {
+    RangeError.checkValidIndex(index, this);
```
lrhn: Consider using 
```dart
RangeError.checkValidIndex(index, this, "index", this.length);
```
to avoid the dynamic `length` read in the check. (For performance).

lib/src/bitarray.dart
```diff
@@ -0,0 +1,317 @@
...
+  int _countBits(int val) {
+    var bits = 0;
+    while (val != 0) {
+      bits += _numOfBitsLookup[val & 0xFF];
+      val >>= 8;
+    }
+    return bits;
+  }
+
+  /// Returns number [true] values in [this].
```
lrhn: Document getters using a noun phrase.

`[true]` and `[this]` are not valid links, they are not source names. Use `true` or `this` instead.

Maybe:
```dart
/// The number of `true` values in this bit list.
```

lib/src/bitarray.dart
```diff
@@ -0,0 +1,317 @@
...
+import 'package:collection/collection.dart' show NonGrowableListMixin;
+
+/// A BitArray implementation for Dart to store boolean values.
+///
+/// Uses list of integers as internal storage to reduce memory usage.
+/// Implemented as fixed-length [List] of [bool].
+/// Throws error when calling any operation with length modifying.
+///
+/// Support several ways to perform logical operations (NOT, OR, AND, XOR):
+/// * instance methods (mutate [this]): [BitArray.not], [BitArray.or], [BitArray.and], [BitArray.xor];
```
lrhn: [this] -> `this`   
(The `this` is not a source name so you cannot link to it).
"mutate" -> "mutates" 
(present tense, not imperative).

lib/src/bitarray.dart
lrhn: This is a bog-standard list iterator which does `[]` for every index up to `length`.  You can just use the one you inherit from `ListMixin`.

lib/src/bitarray.dart
```diff
@@ -0,0 +1,317 @@
...
+    _data.last &= _tailMask;
+  }
+
+  /// Performs logical negation to [this].
+  void not() {
+    _transformDataWords((int word, _) => ~word);
+  }
+
+  /// Performs logical AND operation to [this] with given [BitArray].
+  void and(BitArray ba) {
```
rakudrama: We use proper words for most names.
`ba` should be renamed, e.g. to `other` (to complement `this`), `operand`, `argument`, or in cases where there is one BitArray (not here, there are two when considering `this`), `bitArray`.

lib/src/bitarray.dart
```diff
@@ -0,0 +1,317 @@
...
+    _checkLengthsEqual(this, ba);
+    _transformDataWords(
+        (int word, int wordIndex) => word & ba._data[wordIndex]);
```
rakudrama: I think these higher-order functions are going to be a performance problem (creating closures that might not be fully inlined).

I would write this body as a loop, and as there are several paths that have a loop doing '&', make that loop be a private static method called from all places
The call
```dart
    _transformDataWords(
        (int word, int wordIndex) => word & ba._data[wordIndex]);
```
would then be replace by
```dart
   _bitwiseAnd(this._data, this._data, ba._data);
```
```dart
// [target] can be the same list as [source1] or [source2].
static void _bitwiseAnd(Uint32List target, Uint32List source1, Uint32List source2) {
  assert(target.length <= source1.length && target.length <= source2.length);
  for (int i = 0; i < target.length; i++) {
    target[i] = source1[i] & source2[i];
  }
  // no tailmask required for '&', but could be passed in.
}
```

lib/src/bitarray.dart
```diff
@@ -0,0 +1,317 @@
...
+
+  /// Performs logical XOR operation to [this] with given [BitArray].
+  void xor(BitArray ba) {
+    _checkLengthsEqual(this, ba);
+    _transformDataWords(
+        (int word, int wordIndex) => word ^ ba._data[wordIndex]);
+  }
+
+  /// Creates a new instance of [BitArray] with the same values as in [this].
+  BitArray clone() {
```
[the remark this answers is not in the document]
rakudrama: +1

lib/src/bitarray.dart
rakudrama: dart2js and DDC will not inline this code.

lib/src/bitarray.dart
```diff
@@ -0,0 +1,317 @@
...
+
+import 'dart:collection' show ListMixin;
+import 'dart:typed_data' show Uint32List;
+import 'package:collection/collection.dart' show NonGrowableListMixin;
+
+/// A BitArray implementation for Dart to store boolean values.
+///
+/// Uses list of integers as internal storage to reduce memory usage.
+/// Implemented as fixed-length [List] of [bool].
+/// Throws error when calling any operation with length modifying.
```
rakudrama: Every time we create a fixed-length list class, at some point in the future someone needs the growable version and writes a completely new implementation. Almost all the code is the same, but duplicated.
I would suggest that this library embrace this future and build on a base class that can be used for both fixed-length and growable implementations.

lib/src/bitarray.dart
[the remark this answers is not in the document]
rakudrama: If we pick one, can the other be made visible as a view like `List.asMap`, but modifiable?

rakudrama: It becomes an (slightly) reusable interface when someone implements the growable version.

lib/src/bitarray.dart
```diff
@@ -0,0 +1,317 @@
...
+  ];
+
+  /// Creates a [BitArray] with [length].
+  ///
+  /// All initial values are [false], set [filled] to [true] to make them [true].
+  BitArray(int length, [bool filled = false]) {
+    ArgumentError.checkNotNull(length);
+    RangeError.checkNotNegative(length, 'length');
+
+    _length = length;
```
rakudrama: Use initializers rather than assignment.
If this is difficult, have a private constructor `BitList._(this._length, this._data, this._tailMask);` and make this constructor be a factory constructor.

lib/src/bitarray.dart
```diff
@@ -0,0 +1,317 @@
...
+/// Support several ways to perform logical operations (NOT, OR, AND, XOR):
+/// * instance methods (mutate [this]): [BitArray.not], [BitArray.or], [BitArray.and], [BitArray.xor];
+/// * producing new instance of [BitArray]:
+///     * logical operators: [BitArray.~], [BitArray.|], [BitArray.&], [BitArray.^].
+///     * logical gates (as static methods): [BitArray.notGate], [BitArray.orGate], [BitArray.andGate], [BitArray.xorGate].
+/// All arguments for not-unary logical operations must have the same [length] or [ArgumentError] will be thrown.
+class BitArray with ListMixin<bool>, NonGrowableListMixin<bool> {
+  int _length;
+
+  Uint32List _data;
```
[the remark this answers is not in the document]
AAAbramenko: `Uint32List` was chosen for performance reasons: it requires for 32/8=4 times less iterating loops than `Uint8List` (in the worst case there will be 3 unused bytes, but it is not too much expensive loss).
`Uint32x64List' can't be considered due impossibility to be compiled to js.

lib/src/bitarray.dart
```diff
@@ -0,0 +1,317 @@
...
+
+  /// Mask to set tail bits to zero.
+  int _tailMask;
+
+  static const List<int> _numOfBitsLookup = [
```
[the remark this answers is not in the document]
AAAbramenko: https://github.com/dart-lang/collection/pull/120#discussion_r389263363

lib/src/bitarray.dart
AAAbramenko: Yes, it is. E.g. C++ [std::vector bool ](https://devdocs.io/cpp/utility/bitset/flip) and Java's [BitSet](https://docs.oracle.com/en/java/javase/13/docs/api/java.base/java/util/BitSet.html#flip(int))

lib/src/bitarray.dart
AAAbramenko: Is using Null Safety OK for custom iterators? Returning non-`null` values before and after iteration violates requirements of `current` property of `Iterator<E>`:
> Returns null if the iterator has not yet been moved to the first element, or if the iterator has been moved past the last element of the Iterable.

lib/src/bitarray.dart
```diff
@@ -0,0 +1,317 @@
...
+  /// Replaces integer values in [_data] by applying [func] to each word.
+  void _transformDataWords(int Function(int word, int wordIndex) func) {
+    for (var wordIndex = 0; wordIndex < _data.length; ++wordIndex) {
+      _data[wordIndex] = func(_data[wordIndex], wordIndex);
+    }
+    _clearTailBits();
+  }
+
+  /// Sets all words in [_data] to values retuned from [func].
+  void _setDataWords(int Function(int wordIndex) func) {
```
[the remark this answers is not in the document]
AAAbramenko: OK. This was done to reduce copy-paste for(..){} loop over the words.

lib/src/bitarray.dart
AAAbramenko: Will be removed. This function calls will be replaced with loops.

lib/src/bitarray.dart
AAAbramenko: These functions will be replaced with at place loops.

lib/src/bitarray.dart
AAAbramenko: I want to make it in the following manner (instead of inherited from ListMixin):
```dart
class _BitArrayIterator implements Iterator<bool> {
  bool _current;
  int _pos = 0;

  int _wordIndex = -1;
  int _bitIndex = BitArray._bitsPerEntry - 1;
  int _word;

  int _mask;

  final BitArray _bitArray;

  _BitArrayIterator(this._bitArray);

  @override
  bool get current => _current;

  @override
  bool moveNext() {
    if (_pos == _bitArray.length) {
      _current = null;
      return false;
    }

    if (++_bitIndex == BitArray._bitsPerEntry) {
      _word = _bitArray._data[++_wordIndex];
      _bitIndex = 0;
      _mask = 1;
    } else {
      _mask <<= 1;
    }

    _current = _word & _mask != 0;
    _pos++;
    return true;
  }
}
```

lib/src/bitarray.dart
```diff
@@ -0,0 +1,317 @@
...
+
+  /// Creates a [BitArray] with [length].
+  ///
+  /// All initial values are [false], set [filled] to [true] to make them [true].
+  BitArray(int length, [bool filled = false]) {
+    ArgumentError.checkNotNull(length);
+    RangeError.checkNotNegative(length, 'length');
+
+    _length = length;
+    _data = Uint32List((length >> 5) + 1);
```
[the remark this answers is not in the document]
AAAbramenko: Value of 31 (`_bitsPerEntry - 1`) is used a lot so I think it will be good to add a constant:
```dart
const int _entryMsbIndex = 31;
```

lib/src/bitarray.dart
AAAbramenko: Will be renamed to BitList.

lib/src/bitarray.dart
```diff
@@ -0,0 +1,317 @@
...
+
+  /// Sets tail bits in the last word of [_data] to zero.
+  ///
+  /// Needs to clear last bits after logical operations over [this].
+  void _clearTailBits() {
+    _data.last &= _tailMask;
+  }
+
+  /// Performs logical negation to [this].
+  void not() {
```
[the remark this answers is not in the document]
AAAbramenko: not() will be removed, because bit inversion can be done by calling flip(0).

lib/src/bitarray.dart
```diff
@@ -0,0 +1,317 @@
...
+    _checkLengthsEqual(ba1, ba2);
+    var result = BitArray(ba1.length);
+    result._setDataWords((int wordIndex) {
+      return ba1._data[wordIndex] ^ ba2._data[wordIndex];
+    });
+    return result;
+  }
+
+  /// Returns [true] if bitarrays have the same length and the same values.
+  bool equal(BitArray ba) {
```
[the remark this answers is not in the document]
AAAbramenko: It will be renamed.

AAAbramenko:
After review I decided to:
1. rename to BoolList;
2. remove some methods and operators to simplify class;
3. make class with growable and fixed-length implementations.
---
Some prerequisites: this PR was inspired by BitSet from Java, std::bitset e.t.c. I think Dart needs to have some way of effective storage for bool values not from third-party packages.

mit-mit:
@lrhn can you take another look?

lib/src/boollist.dart
```diff
@@ -0,0 +1,252 @@
...
+
+  factory BoolList._selectType(int length, bool growable) {
+    if (growable) {
+      return _GrowableBoolList(length);
+    } else {
+      return _NonGrowableBoolList(length);
+    }
+  }
+
+  /// Creates a [BoolList] with given length.
```
lrhn, line 41: -> with the given length.

lib/src/boollist.dart
```diff
@@ -0,0 +1,252 @@
...
+  ///
+  /// The created list is fixed-length if [length] is provided.
+  /// Al initial values are `false`.
```
lrhn, line 44: I wouldn't copy the `List` constructor behavior. We removed it for a reason.

I'm slightly torn on whether a growable list is something you'll use often or not. That matters for which constructor should be the default.
I am *leaning* towards making the fixed-length constructor the default, which means that `BoolList` should take a length and default to a fixed-length false-filled list of that length.
Then we can add another constructor for growable lists.
```dart
  /// Creates a list of booleans with the provided length.
  ///
  /// The list is initially filled with the [fill] value, and
  /// the list is growable if [growable] is true.
  BoolList(int length, {bool fill = false, bool growable = false}) ...

  /// Creates an empty list of booleans.
  /// 
  /// The list defaults to being growable unless [growable] is `false`.
  /// If [capacity] is provided, and [growable] is not `false`,
  /// the implementation will attempt to make space for that
  /// many elements before needing to grow its internal storage.
  BoolList.empty({bool growable = true, int capacity = 0}) ...
```

That would replace the current `.filled` and unnamed constructor.

lib/src/boollist.dart
```diff
@@ -0,0 +1,252 @@
...
+    _length = length;
+  }
+
+  void _shrink(int length) {
+    if (length < _length ~/ _growthFactor) {
+      var newDataLength = BoolList._lengthInWords(length);
+      _data = Uint32List(newDataLength)..setRange(0, newDataLength, _data);
+    }
+
+    for (var i = length; i < _data.length * BoolList._bitsPerEntry; i++) {
```
lrhn, line 197: Or
```dart
   _data[_data.length - 1] &= (1 << (length & (_bitsPerEntry - 1))) - 1;
```
to clear the remaining bits of the last word.

lib/src/boollist.dart
```diff
@@ -0,0 +1,252 @@
...
+}
+
+class _BoolListIterator implements Iterator<bool> {
+  bool _current = false;
+  int _pos = 0;
+  final int _length;
+  int _wordIndex = -1;
+  int _bitIndex = BoolList._entrySignBitIndex;
+
+  int _mask;
```
lrhn, line 220: Should be initialized. Just set it to 0 or 1. I know it will be set to `1` in the first moveNext call, but it's better to tell the compiler that it won't ever be `null`.

lib/src/boollist.dart
```diff
@@ -0,0 +1,252 @@
...
+    if (_boolList._length != _length) {
+      throw ConcurrentModificationError(_boolList);
+    }
+
+    if (_pos == _boolList.length) {
+      _current = false;
+      return false;
+    }
+
+    if (++_bitIndex == BoolList._bitsPerEntry) {
```
lrhn, line 240: Not sure the extra fields are worth the overhead of maintaining them.

Maybe just do:
```dart
bool moveNext() {
  if (_pos < _boolList.length) {
    _current = _boolList[_pos++];
    // or, to avoid the extra range check:
    //  var pos = _pos++;
    //  _current = 
    //     _boolList._data[pos >> _entryShift] & (1 << (pos & _entrySignBitIndex)) != 0;
    return true;
  }
  _current = false;
  return false;
}
```

lib/src/boollist.dart
```diff
@@ -0,0 +1,252 @@
+// Copyright (c) 2020, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:collection' show ListMixin;
+import 'dart:typed_data' show Uint32List;
+
+import 'unmodifiable_wrappers.dart' show NonGrowableListMixin;
+
+/// A BoolList implementation for Dart to store boolean values.
```
lrhn, line 10:
```dart
/// A space-efficient list of boolean values.
```

Being nit-picky here :) 
The "for Dart" is redundant in a Dart file. Repeating the class name isn't saying anything new. The first line of a DartDoc is intended to grab the attention of someone looking for the functionality that the class is providing, without going into too much detail.

lib/src/boollist.dart
```diff
@@ -0,0 +1,252 @@
...
+///
+/// Uses list of integers as internal storage to reduce memory usage.
+///
+/// Internal storage of the growable [BoolList] is dynamically changed in the
+/// following cases:
```
lrhn, line 15: I wouldn't document the exact algorithm for growing and shrinking. It's better to be able to tweak the limits without anyone complaining.

lrhn (commented):
Looking very good now!

lib/src/boollist.dart
AAAbramenko: Do we need to keep `.generate` and `.from` constructors?

lrhn: I'd keep them. They're obviously not *necessary*, you can always just create a list of the required length and fill in the data, but they are convenient in some situations. (Now, if we could make bool-list literals, then we probably wouldn't need those two, but we can't do `BoolList[for (var i = 0; i < 10; i++) generator(i)]` or `BoolList[...source]`. Yet!)

lib/src/boollist.dart
AAAbramenko: Unfortunately:
1. This will failed in case where:
- new `length` is 0 (`_data[-1]`);
- `length` equals `_bitsPerEntry * _data.length`, last value in `_data` will be set to 0.
2. Growing/shrinking algorithm not replaces `_data` when new `length` is not less than `_length ~/ _growthFactor`. In this case cleaning bits in the last value of `_data` wrong, we need to clean all bits from the 'length' till last bit.

lrhn: ACK. Edge cases. Doing too much checking ahead of time to avoid the edge cases might end up taking more time than just doing the brute-force thing.

AAAbramenko:
Too mush time passed with this PR. 
What shall we do?
Fix and merge, close and forget about It, or this kind of data structure must be part of another package/library?
---
As I mentioned above idea of this class was to keep array of boolean values with minimal memory usage.

lib/collection.dart
```diff
@@ -4,6 +4,8 @@
 
 export 'src/algorithms.dart'
     show binarySearch, insertionSort, lowerBound, mergeSort, shuffle, reverse;
+export 'src/algorithms.dart';
```
lrhn, line 7: This export should be removed.

lrhn (approved):
Sorry about dropping the ball on this.
It still looks very good. There is one incorrect `export`, but otherwise I think it's fine to land as written.
I'll fix that export and land it. (Then I'll add the necessary changelog entries mysefl).

PR #123, opened 2020-03-16
Null safety migration

jakemac53:
Sending this one out for a review even before creating the official branch just due to the size of it, and the relative complexity of things like the Equality classes.

Note that not all tests are passing right now, the `wrapper_test.dart` is broken but due to violations in the test itself, and the `PriorityQueue` and `QueueList` tests are broken due to what I believe is an sdk issue which I will file and link here once I confirm.

### DIFF
--- a/lib/src/algorithms.dart
+++ b/lib/src/algorithms.dart
@@ -15,7 +15,8 @@ import 'utils.dart';
 /// the objects. If any object is not [Comparable], this throws a [CastError].
 ///
 /// Returns -1 if [value] is not in the list by default.
-int binarySearch<T>(List<T> sortedList, T value, {int Function(T, T) compare}) {
+int binarySearch<T>(List<T> sortedList, T value,
+    {int Function(T, T)? compare}) {
   compare ??= defaultCompare<T>();
   var min = 0;
   var max = sortedList.length;
@@ -44,7 +45,7 @@ int binarySearch<T>(List<T> sortedList, T value, {int Function(T, T) compare}) {
 ///
 /// Returns [sortedList.length] if all the items in [sortedList] compare less
 /// than [value].
-int lowerBound<T>(List<T> sortedList, T value, {int Function(T, T) compare}) {
+int lowerBound<T>(List<T> sortedList, T value, {int Function(T, T)? compare}) {
   compare ??= defaultCompare<T>();
   var min = 0;
   var max = sortedList.length;
@@ -64,7 +65,7 @@ int lowerBound<T>(List<T> sortedList, T value, {int Function(T, T) compare}) {
 /// Shuffles a list randomly.
 ///
 /// A sub-range of a list can be shuffled by providing [start] and [end].
-void shuffle(List list, [int start = 0, int end]) {
+void shuffle(List list, [int start = 0, int? end]) {
   var random = math.Random();
   end ??= list.length;
   var length = end - start;
@@ -78,7 +79,7 @@ void shuffle(List list, [int start = 0, int end]) {
 }
 
 /// Reverses a list, or a part of a list, in-place.
-void reverse(List list, [int start = 0, int end]) {
+void reverse(List list, [int start = 0, int? end]) {
   end ??= list.length;
   _reverse(list, start, end);
 }
@@ -108,7 +109,7 @@ void _reverse(List list, int start, int end) {
 /// This insertion sort is stable: Equal elements end up in the same order
 /// as they started in.
 void insertionSort<T>(List<T> list,
-    {int Function(T, T) compare, int start = 0, int end}) {
+    {int Function(T, T)? compare, int start = 0, int? end}) {
   // If the same method could have both positional and named optional
   // parameters, this should be (list, [start, end], {compare}).
   compare ??= defaultCompare<T>();
@@ -151,7 +152,7 @@ const int _MERGE_SORT_LIMIT = 32;
 /// This merge sort is stable: Equal elements end up in the same order
 /// as they started in.
 void mergeSort<T>(List<T> list,
-    {int start = 0, int end, int Function(T, T) compare}) {
+    {int start = 0, int? end, int Function(T, T)? compare}) {
   end ??= list.length;
   compare ??= defaultCompare<T>();
 
@@ -171,7 +172,7 @@ void mergeSort<T>(List<T> list,
   var firstLength = middle - start;
   var secondLength = end - middle;
   // secondLength is always the same as firstLength, or one greater.
-  var scratchSpace = List<T>(secondLength);
+  var scratchSpace = List<T?>(secondLength);
   _mergeSort(list, compare, middle, end, scratchSpace, 0);
   var firstTarget = end - firstLength;
   _mergeSort(list, compare, start, middle, list, firstTarget);
@@ -184,7 +185,7 @@ void mergeSort<T>(List<T> list,
 ///
 /// It will work in-place as well.
 void _movingInsertionSort<T>(List<T> list, int Function(T, T) compare,
-    int start, int end, List<T> target, int targetOffset) {
+    int start, int end, List<T?> target, int targetOffset) {
   var length = end - start;
   if (length == 0) return;
   target[targetOffset] = list[start];
@@ -194,7 +195,7 @@ void _movingInsertionSort<T>(List<T> list, int Function(T, T) compare,
     var max = targetOffset + i;
     while (min < max) {
       var mid = min + ((max - min) >> 1);
-      if (compare(element, target[mid]) < 0) {
+      if (compare(element, target[mid]!) < 0) {
         max = mid;
       } else {
         min = mid + 1;
@@ -213,7 +214,7 @@ void _movingInsertionSort<T>(List<T> list, int Function(T, T) compare,
 /// Allows target to be the same list as [list], as long as it's not
 /// overlapping the `start..end` range.
 void _mergeSort<T>(List<T> list, int Function(T, T) compare, int start, int end,
-    List<T> target, int targetOffset) {
+    List<T?> target, int targetOffset) {
   var length = end - start;
   if (length < _MERGE_SORT_LIMIT) {
     _movingInsertionSort(list, compare, start, end, target, targetOffset);
@@ -246,10 +247,10 @@ void _merge<T>(
     List<T> firstList,
     int firstStart,
     int firstEnd,
-    List<T> secondList,
+    List<T?> secondList,
     int secondStart,
     int secondEnd,
-    List<T> target,
+    List<T?> target,
     int targetOffset) {
   // No empty lists reaches here.
   assert(firstStart < firstEnd);
@@ -257,7 +258,7 @@ void _merge<T>(
   var cursor1 = firstStart;
   var cursor2 = secondStart;
   var firstElement = firstList[cursor1++];
-  var secondElement = secondList[cursor2++];
+  var secondElement = secondList[cursor2++]!;
   while (true) {
     if (compare(firstElement, secondElement) <= 0) {
       target[targetOffset++] = firstElement;
--- a/lib/src/algorithms.dart
+++ b/lib/src/algorithms.dart
@@ -267,7 +267,7 @@ void _merge<T>(
     } else {
       target[targetOffset++] = secondElement;
       if (cursor2 != secondEnd) {
-        secondElement = secondList[cursor2++];
+        secondElement = secondList[cursor2++]!;
         continue;
       }
       // Second list empties first. Flushing first list here.
--- a/lib/src/algorithms.dart
+++ b/lib/src/algorithms.dart
@@ -172,7 +172,7 @@ void mergeSort<T>(List<T> list,
   var firstLength = middle - start;
   var secondLength = end - middle;
   // secondLength is always the same as firstLength, or one greater.
-  var scratchSpace = List<T?>(secondLength);
+  var scratchSpace = List<T?>.filled(secondLength, null);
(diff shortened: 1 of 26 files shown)


lib/src/algorithms.dart
```diff
@@ -171,7 +172,7 @@ void mergeSort<T>(List<T> list,
   var firstLength = middle - start;
   var secondLength = end - middle;
   // secondLength is always the same as firstLength, or one greater.
-  var scratchSpace = List<T>(secondLength);
+  var scratchSpace = List<T?>(secondLength);
```
natebosch: I thought this constructor was going away and we'd have to use `List<T?>.filled(secondLength, null)`

jakemac53: It works in the current SDK - removing that entirely would break huge swaths of existing code? I don't think we have a way to only remove it in the nnbd sdk...

natebosch: https://github.com/dart-lang/language/issues/746#issuecomment-571377956

cc @leafpetersen - is there a tracking issue for implementing the error?

lib/src/canonicalized_map.dart
```diff
@@ -30,8 +30,8 @@ class CanonicalizedMap<C, K, V> implements Map<K, V> {
   /// The [isValidKey] function is called before calling [canonicalize] for
   /// methods that take arbitrary objects. It can be used to filter out keys
   /// that can't be canonicalized.
-  CanonicalizedMap(C Function(K key) canonicalize,
-      {bool Function(Object key) isValidKey})
+  CanonicalizedMap(C Function(K? key) canonicalize,
```
natebosch: This is unfortunate since it makes everyone using it have to handle nulls even if they do't want to. The `K` could already be nullable, why do we need it to be nullable here?

lib/src/combined_wrappers/combined_iterable.dart
```diff
@@ -45,17 +45,30 @@ class _CombinedIterator<T> implements Iterator<T> {
   /// avoid instantiating unnecessary iterators.
   final Iterator<Iterator<T>> _iterators;
 
-  _CombinedIterator(this._iterators);
+  /// The current iterator in [_iterators], or `null` if done iterating.
```
natebosch: I don't quite follow why we need this. Why did we have to `_advance()` in the constructor?

lib/src/algorithms.dart
jakemac53: Fwiw the migrator tool also does not migrate this yet as it was indidcated it should in that issue. That would alleviate some of my concern here. cc @stereotype441

lib/src/functions.dart
```diff
@@ -61,12 +61,14 @@ Map<T, List<S>> groupBy<S, T>(Iterable<S> values, T Function(S) key) {
 /// The values returned by [orderBy] are compared using the [compare] function.
 /// If [compare] is omitted, values must implement [Comparable<T>] and they are
 /// compared using their [Comparable.compareTo].
-S minBy<S, T>(Iterable<S> values, T Function(S) orderBy,
-    {int Function(T, T) compare}) {
+///
+/// Returns `null` if [values] is empty.
```
natebosch: We should consider whether an exception would be better.

We should also consider if we should drop the top level method in favor of an extension.

lib/src/priority_queue.dart
```diff
@@ -128,7 +129,7 @@ class HeapPriorityQueue<E> implements PriorityQueue<E> {
   final Comparator<E> comparison;
 
   /// List implementation of a heap.
-  List<E> _queue = List<E>(_INITIAL_CAPACITY);
+  List<E?> _queue = List<E?>(_INITIAL_CAPACITY);
```
natebosch:
```suggestion
  List<E?> _queue = List<E?>.filled(_INITIAL_CAPACITY, null);
```

jakemac53:
SDK issue is here https://github.com/dart-lang/sdk/issues/41060

lib/src/priority_queue.dart
```diff
@@ -222,19 +223,14 @@ class HeapPriorityQueue<E> implements PriorityQueue<E> {
...
   @override
-  List<E> toList() {
-    var list = <E>[]
-      ..length = _length
-      ..setRange(0, _length, _queue)
-      ..sort(comparison);
-    return list;
-  }
+  List<E> toList() =>
+      List<E>.generate(_length, (i) => _queue[i]!)..sort(comparison);
```
natebosch: [optional]
```suggestion
      [for (var i = 0; i < _length; i++) _qeue[i]!]..sort(comparison);
```

lib/src/queue_list.dart
```diff
@@ -261,7 +261,9 @@ class QueueList<E> extends Object with ListMixin<E> implements Queue<E> {
 class _CastQueueList<S, T> extends QueueList<T> {
   final QueueList<S> _delegate;
 
-  _CastQueueList(this._delegate) : super._() {
+  /// Assigns invalid values for head/tail because it uses the delegate to hold
```
natebosch: [nit] Should this be a `//` comment?

lib/src/queue_list.dart
```diff
@@ -261,7 +261,9 @@ class QueueList<E> extends Object with ListMixin<E> implements Queue<E> {
...
+  /// the real values, but they are non-null fields.
+  _CastQueueList(this._delegate) : super._(-1, -1) {
     _table = _delegate._table.cast<T>();
   }
```
natebosch: I'm pretty sure we can access `_delegate` in the initializer list now.
```suggestion
  _CastQueueList(this._delegate)
      : _table = _delegate._table.cast<T>(),
        super._(-1, -1);
```

lib/src/union_set_controller.dart
```diff
@@ -23,8 +23,8 @@ import 'union_set.dart';
 /// ```
 class UnionSetController<E> {
   /// The [UnionSet] that provides a view of the union of sets in [this].
-  UnionSet<E> get set => _set;
-  UnionSet<E> _set;
+  UnionSet<E>? get set => _set;
+  UnionSet<E>? _set;
```
natebosch: We don't want these to be nullable. Maybe this should be `late`?

Can we move the assignment to an initializer list and make if `final` and non-nullable?

lib/src/union_set_controller.dart
```diff
@@ -50,5 +50,5 @@ class UnionSetController<E> {
   ///
   /// If another set in [this] has overlapping elements with [component], those
   /// elements will remain in [set].
-  bool remove(Set<E> component) => _sets.remove(component);
+  bool remove(Set<E>? component) => _sets.remove(component);
```
natebosch: `component` should not be nullable

lib/src/wrappers.dart
```diff
@@ -200,7 +200,7 @@ class DelegatingList<E> extends DelegatingIterable<E> implements List<E> {
   }
 
   @override
-  void fillRange(int start, int end, [E fillValue]) {
+  void fillRange(int start, int end, [E? fillValue]) {
```
natebosch: Did this signature come from the SDK? We should maybe file an issue, I don't know why we'd accept a nullable `fillValue` here. I guess because we don't have a default?

test/algorithms_test.dart
```diff
@@ -212,14 +212,12 @@ void main() {
...
-      var list = List<OC>(size);
       // Class OC compares using id.
       // With size elements with id's in the range 0..size/4, a number of
       // collisions are guaranteed. These should be sorted so that the 'order'
       // part of the objects are still in order.
-      for (var i = 0; i < size; i++) {
-        list[i] = OC(random.nextInt(size >> 2), i);
-      }
+      var list =
+          List<OC>.generate(size, (i) => OC(random.nextInt(size >> 2), i));
```
natebosch: [optional] for loop element might look nicer here.

```suggestion
          [for (var i = 0; i < size; i++ ) OC(random.nextInt(size >> 2), i)];
```

lib/src/combined_wrappers/combined_iterable.dart
jakemac53: You are always supposed to call `moveNext()` before asking for `current`. The behavior is undefined and we previously relied on the fact that it would return `null`. This is no longer possible to rely on with NNBD, and its better not to anyways :). Now we properly `moveNext` instead.

lib/src/canonicalized_map.dart
jakemac53: Because of this comment on the class:

```
/// By default, `null` is allowed as a key. It can be forbidden via the
/// `isValidKey` parameter.
```

We could change that behavior though potentially?

natebosch: We should consider it. The UX is likely a lot worse this way. I would guess a lot of usages in the wild don't bother handling null keys because they just never come up.

jakemac53: It was pretty explicitly handled here too https://github.com/dart-lang/collection/blob/master/lib/src/canonicalized_map.dart#L182.

But, we could just remove the explicit null check there and only leave the `is K` check.

This would be a more breaking change I think though.

lib/src/algorithms.dart
jakemac53: I have updated everything to use `List.filled` now though.

lib/src/canonicalized_map.dart
natebosch: I think it was handled there because `null is K` behaved weirdly before.

With null safety you can always make `K` nullable if you want that behavior.

jakemac53: `null is T` would have always been false before (unless T was Null) afaik? It does seem like it was an intentional choice to me but ya I don't really understand why.

natebosch: `null as T` would have worked though. I think it was an because `null` was always allowed as an argument for any `T`.

lib/src/functions.dart
jakemac53: I would prefer to do things like that after the NNBD migration and do a non-breaking change in this package.

natebosch: We should sync on this plan. I really don't like the idea of handcuffing ourselves in our API decisions because of a bad practice. The SDK made breaking changes in order to improve null safe APIs, I don't see why we shouldn't do the same in packages.

jakemac53: The sdk is significantly more hamstrung than we are with regards to versioning. There isn't anything preventing us from doing a non-breaking nnbd release followed by a breaking release.

Most users would be able to support both releases I think as well.

natebosch: This already _is_ breaking, we are already lying if we release it as non-breaking.

natebosch: I would be fine with a deprecation period, either before or after the null safety release.

jakemac53: Do you mean its breaking outside of the out of order migration case or is that what you are referring to?

jakemac53: If we deprecate it we could do leave this as is, and the new api would have the non-nullable return type with the throw.

natebosch: I mean the out of order migration case.

A deprecation may work for this case since we can replace the API, but won't work in all cases.

jakemac53: We can always push additional commits to this branch for breaking changes. At least for this initial PR I would prefer to do a clean migration.

lib/src/priority_queue.dart
jakemac53: leaving as is since it is closer to the original behavior (its theoretically faster since it specifies the length)

lib/src/queue_list.dart
jakemac53: done

lib/src/queue_list.dart
jakemac53: Unfortunately we can't - `_table` is actually from the super class (🤢)

lib/src/union_set_controller.dart
jakemac53: I made it late final - I can't move the assignment to the initializer list because it requires another field.

I could make a factory and private constructor if you think that would be better though...

lib/src/union_set_controller.dart
jakemac53: done

lib/src/wrappers.dart
jakemac53: Ya, https://github.com/dart-lang/sdk/blob/master/sdk_nnbd/lib/core/list.dart#L716.

I agree its weird, but yes I assume its because we couldn't make it a required parameter and not break other users. I wonder if we just used the `required` keyword though?

lib/src/queue_list.dart
natebosch: We can pass it through the `super._` call then. I suspect we could even make it final and initialize it in the initializer list in the parent class. If you want to file an issue I can follow up on this outside of this PR.

lib/src/union_set_controller.dart
natebosch: factory + private constructor would, I think, be technically breaking if someone `extends UnionSet`, but we maybe can ignore that. Let's file an issue and follow up separately.

lib/src/wrappers.dart
natebosch: I think `required` only works for named args. This unfortunate but I don't know if we have much choice. 😢

natebosch (approved):

lib/src/union_set_controller.dart
jakemac53: https://github.com/dart-lang/collection/issues/124

lib/src/queue_list.dart
jakemac53: done

PR #125, opened 2020-03-18
Replace a use of putIfAbsent with ??=

whesse:
Because it avoids using a closure, it seems that ??= [] can be used with better performance than putIfAbsent in groupBy.


kevmoo:
Sad this is the case, but it's what @rakudrama has told me, too!

Issue #127, opened 2020-03-23
wrapper_test.dart is fundamentally incompatible with NNBD


jakemac53:
Specifically it uses `noSuchMethod` to intercept function calls and _always_ returns null, which doesn't work if the return type of the method isn't nullable.

jakemac53:
I resolved this with some mirrors hacks

PR #128, opened 2020-03-23
update deps and fix tests so they run with --null-safety

jakemac53:
Note that wrapper_test.dart was not migrated as it is incompatible, https://github.com/dart-lang/collection/issues/127.

### DIFF
--- a/lib/src/algorithms.dart
+++ b/lib/src/algorithms.dart
@@ -176,8 +176,8 @@ void mergeSort<T>(List<T> list,
   _mergeSort<T>(list, compare, middle, end, scratchSpace, 0);
   var firstTarget = end - firstLength;
   _mergeSort<T>(list, compare, start, middle, list, firstTarget);
-  _merge<T>(compare, list, firstTarget, end, scratchSpace, 0, secondLength, list,
-      start);
+  _merge<T>(compare, list, firstTarget, end, scratchSpace.cast<T>(), 0,
+      secondLength, list, start);
 }
 
 /// Performs an insertion sort into a potentially different list than the
@@ -230,8 +230,8 @@ void _mergeSort<T>(List<T> list, int Function(T, T) compare, int start, int end,
   // Sort the first half into the end of the source area.
   _mergeSort<T>(list, compare, start, middle, list, middle);
   // Merge the two parts into the target area.
-  _merge<T>(compare, list, middle, middle + firstLength, target, targetMiddle,
-      targetMiddle + secondLength, target, targetOffset);
+  _merge<T>(compare, list, middle, middle + firstLength, target.cast<T>(),
+      targetMiddle, targetMiddle + secondLength, target, targetOffset);
 }
 
 /// Merges two lists into a target list.
@@ -247,7 +247,7 @@ void _merge<T>(
     List<T> firstList,
     int firstStart,
     int firstEnd,
-    List<T?> secondList,
+    List<T> secondList,
     int secondStart,
     int secondEnd,
     List<T?> target,
--- a/lib/src/queue_list.dart
+++ b/lib/src/queue_list.dart
@@ -127,7 +127,7 @@ class QueueList<E> extends Object with ListMixin<E> implements Queue<E> {
   @override
   E removeFirst() {
     if (_head == _tail) throw StateError('No element');
-    var result = _table[_head]!;
+    var result = _table[_head] as E;
     _table[_head] = null;
     _head = (_head + 1) & (_table.length - 1);
     return result;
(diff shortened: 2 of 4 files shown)


lib/src/algorithms.dart
```diff
@@ -247,7 +247,7 @@ void _merge<T>(
     List<T> firstList,
     int firstStart,
     int firstEnd,
-    List<T?> secondList,
+    List<T> secondList,
```
natebosch: I worry the `cast` is expensive. Why did this need to change?

lib/src/queue_list.dart
```diff
@@ -150,6 +150,15 @@ class QueueList<E> extends Object with ListMixin<E> implements Queue<E> {
   @override
   set length(int value) {
     if (value < 0) throw RangeError('Length $value may not be negative.');
+    if (value > length) {
+      try {
+        null as E;
+      } on CastError {
```
natebosch: We might need to double check behavior on dart2js with `-o3`

cc @sigmundch

test/queue_list_test.dart
```diff
@@ -136,14 +136,18 @@ void main() {
...
+      var queue = QueueList<int?>.from([1, 2, 3]);
       queue.length = 5;
       expect(queue, equals([1, 2, 3, null, null]));
     });
 
     test('throws a RangeError if length is less than 0', () {
       expect(() => QueueList().length = -1, throwsRangeError);
     });
+
+    test('throws an UnsupportedError if element type is non-nullable', () {
```
natebosch: We can _only_ run this test in strong mode. Do we care? Is there a reason we might want to run tests in weak mode too for some time?

lib/src/algorithms.dart
jakemac53: @lrhn do you want to take a look at this?

The problem is that `_merge` calls `target.setRange` with this List. The `target` parameter has a type of `List<T?>` but we actually pass in a `List<T>`, so the `setRange` call fails.

I am not sure what the most efficient way to fix this code is going to be, this version adds a lot of extra casting.

I think for now we can just land this fix but it would be good to optimize it.

test/queue_list_test.dart
jakemac53: We can cross that bridge when we come to it imo

lib/src/queue_list.dart
sigmundch: /cc @rakudrama 

I believe in this scenario you are safe, even with -O3, but I worry in general when I see a `catch` for CastError. BTW, I believe in weak-mode this is trivially true (we always allow null, even on non-nullable types)

An alternative w/o try-catch could be to do:
```dart
void Function(T) _foo<T>() => (T a) {};
bool _isNullable<T>() => _foo<T>() is void Function(Null); // True if Null <: T
...
   if (value > length) {
      if (_isNullable<E>()) throw UnsupportedError(...);
```

dart2js may optimize away the cast as trivially true in weak mode, but not in strong mode. The function subtype check is unlikely to be optimized away, but because it doesn't involve try-catch blocks it may end up smaller in strong-mode.

If detecting nullability of a type-parameter is a common pattern, maybe we could expose a common API for it? Or specialize it differently in the compiler?

nit: if you keep the current approach, you should catch `TypeError`, since `CastError` is now deprecated (currently TypeError implements CastError). I'd also recommend creating a helper method in case you end up needing it in a few places, and use `@pragma('dart2js:tryInline')`

lib/src/algorithms.dart
```diff
@@ -176,8 +176,8 @@ void mergeSort<T>(List<T> list,
   _mergeSort<T>(list, compare, middle, end, scratchSpace, 0);
   var firstTarget = end - firstLength;
   _mergeSort<T>(list, compare, start, middle, list, firstTarget);
-  _merge<T>(compare, list, firstTarget, end, scratchSpace, 0, secondLength, list,
-      start);
+  _merge<T>(compare, list, firstTarget, end, scratchSpace.cast<T>(), 0,
```
lrhn: I would initialize scratchSpace to `List<T>.filled(secondLength, list[start]);`.

Since we know that  0 &le; `start` &lt; `start` + `_MERGE_SORT_LIMIT` &le; `end` &le; `list.length`, that value exist.
(Or we would have known if the function had started with the appropriate:
```dart
   end = RangeError.checkValidRange(start, end, list.length);
```
so consider adding that while you are here).

I'd make the parameter in `_mergeSort` and `_merge` below a `List<T>` as well.

lib/src/queue_list.dart
lrhn: Agree. Always use `TypeError`, never use `CastError` ever again. There is no `CastError` (eventually).

test/queue_list_test.dart
lrhn: That bridge needs to be crossed before landing, or Travis will probably yell at you.

lib/src/algorithms.dart
jakemac53: done - that worked thanks!

lib/src/queue_list.dart
jakemac53: ack

PR #135, opened 2020-05-13
Add extension methods to package:collection.

lrhn:
Add a number of useful extension methods to `List` and `Iterable`, and to a few other types.

A good part of these extensions are backed by the algorithms of `algorithms.dart`, which is expanded to support them:

* Added `quickSort`.
* Add extra optional `Random` argument to `shuffle`.
* Generalize some of the functions in `algorithms.dart` to work on specific properties of the objects (`binarySearchBy`,
  `lowerBoundBy`, `insertionSortBy`, `quickSortBy`, `mergeSortBy`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,20 @@
 * Remove the unusable setter `UnionSetController.set=`. This was mistakenly
   added to the public API but could never be called.
 
+* Added `quickSort` to `algorithms.dart`.
+
+* Add extra optional `Random` argument to `shuffle`.
+
+* Generalize some of the functions in `algorithms.dart` to
+  work on specific properties of the objects (`binarySearchBy`,
+  `lowerBoundBy`, `insertionSortBy`, `quickSortBy`, `mergeSortBy`).
+
+* Add a large number of extension methods on `Iterable` and `List` types,
+  and on a few other types.
+  These either provide easy access to the operations from `algorithms.dart`,
+  or provide convenience variants of existing `Iterable` and `List` methods
+  like `singleWhereOrNull` or `forEachIndexed`.
+
 ## 1.15.0-nullsafety.3
 
 * Allow 2.10 stable and 2.11.0 dev SDK versions.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,13 +3,7 @@
 * Remove the unusable setter `UnionSetController.set=`. This was mistakenly
   added to the public API but could never be called.
 
-* Added `quickSort` to `algorithms.dart`.
-
 * Add extra optional `Random` argument to `shuffle`.
-
-* Generalize some of the functions in `algorithms.dart` to
-  work on specific properties of the objects (`binarySearchBy`,
-  `lowerBoundBy`, `insertionSortBy`, `quickSortBy`, `mergeSortBy`).
 
 * Add a large number of extension methods on `Iterable` and `List` types,
   and on a few other types.
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -1,4 +1,4 @@
-include: package:pedantic/analysis_options.yaml
+include: package:pedantic/analysis_options.1.9.0.yaml
 analyzer:
   errors:
     unused_element: error
@@ -7,39 +7,21 @@ analyzer:
     dead_code: error
   enable-experiment:
     - non-nullable
-
 linter:
   rules:
     # Errors
-    - avoid_empty_else
     - control_flow_in_finally
     - empty_statements
     - hash_and_equals
     - implementation_imports
     - test_types_in_equals
     - throw_in_finally
-    - unrelated_type_equality_checks
-    - valid_regexps
 
     # Style
-    - avoid_init_to_null
     - avoid_private_typedef_functions
     - avoid_renaming_method_parameters
-    - avoid_return_types_on_setters
     - await_only_futures
     - camel_case_types
     - directives_ordering
-    - empty_catches
-    - empty_constructor_bodies
-    - library_names
-    - library_prefixes
     - non_constant_identifier_names
     - only_throw_errors
-    - prefer_equal_for_default_values
-    - prefer_final_fields
-    - prefer_generic_function_type_aliases
-    - prefer_is_not_empty
-    - slash_for_doc_comments
-    - type_init_formals
-    - unnecessary_const
-    - unnecessary_new
--- a/lib/src/algorithms.dart
+++ b/lib/src/algorithms.dart
@@ -2,7 +2,10 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-import 'dart:math' as math;
+/// A selection of data manipulation algorithms.
+library pkg.collection.algorithms;
+
+import 'dart:math' show Random;
 
 import 'utils.dart';
 
@@ -12,19 +15,35 @@ import 'utils.dart';
 /// is unpredictable.
 ///
 /// If [compare] is omitted, this defaults to calling [Comparable.compareTo] on
-/// the objects. If any object is not [Comparable], this throws a [TypeError]
-/// (`CastError` on some SDK versions).
+/// the objects. In this case, the objects must be [Comparable].
+///
+/// Returns -1 if [value] is not in the list.
+int binarySearch<E>(List<E> sortedList, E value,
+    {int Function(E, E)? compare}) {
+  compare ??= defaultCompare;
+  return binarySearchBy<E, E>(sortedList, identity, compare, value);
+}
+
+/// Returns a position of the [value] in [sortedList], if it is there.
+///
+/// If the list isn't sorted according to the [compare] function on the [keyOf]
+/// property of the elements, the result is unpredictable.
 ///
 /// Returns -1 if [value] is not in the list by default.
-int binarySearch<T>(List<T> sortedList, T value,
-    {int Function(T, T)? compare}) {
-  compare ??= defaultCompare();
-  var min = 0;
-  var max = sortedList.length;
+///
+/// If [start] and [end] are supplied, only that range is searched,
+/// and only that range need to be sorted.
+int binarySearchBy<E, K>(List<E> sortedList, K Function(E element) keyOf,
+    int Function(K, K) compare, E value,
+    [int start = 0, int? end]) {
+  end = RangeError.checkValidRange(start, end, sortedList.length);
+  var min = start;
+  var max = end;
+  var key = keyOf(value);
   while (min < max) {
     var mid = min + ((max - min) >> 1);
     var element = sortedList[mid];
-    var comp = compare(element, value);
+    var comp = compare(keyOf(element), key);
     if (comp == 0) return mid;
     if (comp < 0) {
       min = mid + 1;
@@ -42,19 +61,36 @@ int binarySearch<T>(List<T> sortedList, T value,
 /// is unpredictable.
 ///
 /// If [compare] is omitted, this defaults to calling [Comparable.compareTo] on
-/// the objects. If any object is not [Comparable], this throws a [TypeError]
-/// (`CastError` on some SDK versions).
+/// the objects. In this case, the objects must be [Comparable].
 ///
 /// Returns [sortedList.length] if all the items in [sortedList] compare less
 /// than [value].
-int lowerBound<T>(List<T> sortedList, T value, {int Function(T, T)? compare}) {
-  compare ??= defaultCompare();
-  var min = 0;
-  var max = sortedList.length;
+int lowerBound<E>(List<E> sortedList, E value, {int Function(E, E)? compare}) {
+  compare ??= defaultCompare;
+  return lowerBoundBy<E, E>(sortedList, identity, compare, value);
+}
+
+/// Returns the first position in [sortedList] that is not before [value].
+///
+/// Elements are compared using the [compare] function of the [keyOf] property of
+/// the elements.
+/// If the list isn't sorted according to this order, the result is unpredictable.
+///
+/// Returns [sortedList.length] if all the items in [sortedList] are before [value].
+///
+/// If [start] and [end] are supplied, only that range is searched,
+/// and only that range need to be sorted.
+int lowerBoundBy<E, K>(List<E> sortedList, K Function(E element) keyOf,
+    int Function(K, K) compare, E value,
+    [int start = 0, int? end]) {
+  end = RangeError.checkValidRange(start, end, sortedList.length);
+  var min = start;
+  var max = end;
+  var key = keyOf(value);
   while (min < max) {
     var mid = min + ((max - min) >> 1);
     var element = sortedList[mid];
-    var comp = compare(element, value);
+    var comp = compare(keyOf(element), key);
     if (comp < 0) {
       min = mid + 1;
     } else {
@@ -67,31 +103,36 @@ int lowerBound<T>(List<T> sortedList, T value, {int Function(T, T)? compare}) {
 /// Shuffles a list randomly.
 ///
 /// A sub-range of a list can be shuffled by providing [start] and [end].
-void shuffle(List list, [int start = 0, int? end]) {
-  var random = math.Random();
-  end ??= list.length;
+///
+/// If [start] or [end] are omitted,
+/// they default to the start and end of the list.
+///
+/// If [random] is omitted, it defaults to a new instance of [Random].
+void shuffle(List elements, [int start = 0, int? end, Random? random]) {
+  random ??= Random();
+  end ??= elements.length;
   var length = end - start;
   while (length > 1) {
     var pos = random.nextInt(length);
     length--;
-    var tmp1 = list[start + pos];
-    list[start + pos] = list[start + length];
-    list[start + length] = tmp1;
+    var tmp1 = elements[start + pos];
+    elements[start + pos] = elements[start + length];
+    elements[start + length] = tmp1;
   }
 }
 
 /// Reverses a list, or a part of a list, in-place.
-void reverse(List list, [int start = 0, int? end]) {
-  end ??= list.length;
-  _reverse(list, start, end);
+void reverse<E>(List<E> elements, [int start = 0, int? end]) {
+  end = RangeError.checkValidRange(start, end, elements.length);
+  _reverse<E>(elements, start, end);
 }
 
 /// Internal helper function that assumes valid arguments.
-void _reverse(List list, int start, int end) {
+void _reverse<E>(List<E> elements, int start, int end) {
   for (var i = start, j = end - 1; i < j; i++, j--) {
-    var tmp = list[i];
-    list[i] = list[j];
-    list[j] = tmp;
+    var tmp = elements[i];
+    elements[i] = elements[j];
+    elements[j] = tmp;
   }
 }
 
@@ -99,8 +140,7 @@ void _reverse(List list, int start, int end) {
 /// insertion sort.
 ///
 /// If [compare] is omitted, this defaults to calling [Comparable.compareTo] on
-/// the objects. If any object is not [Comparable], this throws a [TypeError]
-/// (`CastError` on some SDK versions).
+/// the objects. In this case, the objects must be [Comparable].
 ///
 /// Insertion sort is a simple sorting algorithm. For `n` elements it does on
 /// the order of `n * log(n)` comparisons but up to `n` squared moves. The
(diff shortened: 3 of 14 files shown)


lrhn:
This is a first crack at adding extension methods to package:collection.

They are not *tested*. Some are probably not useful (I expect `Iterable<T extends num>.sum()` to be mostly speculative).

analysis_options.yaml
```diff
@@ -1,42 +1,6 @@
...
-    - only_throw_errors
-    - prefer_equal_for_default_values
-    - prefer_final_fields
-    - prefer_generic_function_type_aliases
-    - prefer_is_not_empty
-    - slash_for_doc_comments
-    - type_init_formals
-    - unnecessary_const
-    - unnecessary_new
+    annotate_overrides: ignore
```
natebosch, line 4: We should be following this lint.

analysis_options.yaml
```diff
@@ -1,42 +1,6 @@
...
+    prefer_single_quotes: ignore
+    use_function_type_syntax_for_parameters: ignore
```
natebosch, line 6: We should be following this lint.

Some of the removed ones that aren't in `pedantic` yet are also useful. We rarely violate them on the team, but I don't see any reason for something like `camel_case_types` or `library_prefixes` to need to wait for code review if we do get a contribution from outside the team.

lib/src/empty_unmodifiable_set.dart
```diff
@@ -6,15 +6,10 @@ import 'dart:collection';
 
 import 'unmodifiable_wrappers.dart';
 
-// Unfortunately, we can't use UnmodifiableSetMixin here, since const classes
-// can't use mixins.
 /// An unmodifiable, empty set that can have a const constructor.
 class EmptyUnmodifiableSet<E> extends IterableBase<E>
+    with UnmodifiableSetMixin<E>
```
natebosch, line 11: This seems like a good change, but I don't understand how it's related. Can this go out in a separate commit?

lib/src/algorithms.dart
```diff
@@ -16,14 +19,24 @@ import 'utils.dart';
...
+    binarySearchBy<E, E>(sortedList, _id, compare ?? defaultCompare<E>(), value);
+
+
+/// Returns a position of the [value] in [sortedList], if it is there.
+///
+/// If the list isn't sorted according to the [compare] function on the [keyOf]
+/// property of the elements, the result is unpredictable.
+///
+/// Returns -1 if [value] is not in the list by default.
+int binarySearchBy<E, K>(List<E> sortedList, K keyOf(E element), int Function(K, K) compare, E value) {
```
natebosch, line 32: I don't think it's a good idea to add new top level methods to this package. Anything new should be added as an extension only. If we want to have these as top level methods internally for some reason, we should be hiding them from our exports.

natebosch (changes requested):
Did something get lost in the upload? There are no extension methods added here. This does look like it might be some work towards adding extensions, but it doesn't match the PR description as is.

lib/src/algorithms.dart
lrhn: I completely disagree with not adding top-level functions.

Top-level functions are still a perfectly good way to introduce general reusable functionality.

Extension methods are less generally useful than top-level functions, but more convenient and discoverable.
That's why I have made the distinction of adding convenient and discoverable APIs using extension methods, and more rough and verbose reusable APIs as top-level functions.

lrhn:
Whoops. Forgot to git-add the new files.

lib/src/empty_unmodifiable_set.dart
lrhn: It's an unrelated drive-by fix. I can do that by itself too.

analysis_options.yaml
lrhn: As you may  be aware, I disagree with following that lint. It's not part of the Dart style guide, and I actually prefer the inline notation for function arguments, both for reading and writing.
These are the three exceptions that I apply because I consistently write code that does not follow them.

I'm using pedantic+modifications because it's convenient (there are good things in there too, just not all of it), not because I think it's what we actually want as the Dart recommended lint set outside of Google3.
(Sure, I can add camel_case_types, but that has never been an issue.)

jakemac53: How is it possibly a good thing to have every random person who touches a codebase changing the lints to fit their personal preference?

natebosch: > It's not part of the Dart style guide, and I actually prefer the inline notation for function arguments, both for reading and writing.

https://dart.dev/guides/language/effective-dart/design#consider-using-function-type-syntax-for-parameters

It is a "Consider" it think for largely historical reasons. In future revisions it's likely to become stronger. The fact that there _are_ 2 syntaxes for this will hopefully be fixed in Dart 3 or some 2.x language version - ideally we have 1 way to do things instead of 2 competing ways that require constant negotiation or inconsistent codebases.

> These are the three exceptions that I apply because I consistently write code that does not follow them.

This goes against the flow of the rest of Google Dart authors, and against the general philosophy of Readability in Google.

> Sure, I can add camel_case_types, but that has never been an issue.

It has never been a _big_ issue. It has come up in code reviews in the past and the intention of putting it in a lint is to save time, and save from re-negotiating it with new contributors. The reason `camel_case_types` isn't in `pedantic` is because that would require consistent enforcement, and in internal code I think we have some exceptions. We won't have exceptions in _this_ package which is why the lint is enabled.


I know that we have different approaches to linting in different places. I don't think having us each try to assert our own style into shared packages is the right thing to do, neither is introducing config changes into other PRs.

lib/src/empty_unmodifiable_set.dart
natebosch: > I can do that by itself too.

I think that would be better. The travis wait time for individual PRs is the only downside - if you need to send a few things together to work around that I think it's OK, otherwise for isolated refactorings I think we should follow go/small-cls

lib/src/algorithms.dart
natebosch: > Not every function deserves looking like an instance-method.

When I last looked it seemed like every function in _this_ packages deserves looking like an instance-method. If we wont' be adding `list.binarySearchBy` extension at all, that does change the discussion a bit, but I think that would be a good extension method.

> It also avoids the "all functions after 2.6 are extension methods, all functions before are not" issue.

I don't think that is an issue, I haven't seen anyone make that mistake.

The main thing I want to push on is to have 1 (recommended) way to do things. I want the users of this package to avoid negotiating during _their_ code review about whether to use the extension method version or the top-level method version.

> The general function doesn't really _fit in_ as an instance/extension member, but it's there in case you really need the full power.

I don't understand this. Is this ability to be passed as a function (with a _super_ specific function signature) rather than be torn off a specific list the "full power"?

lib/src/iterable_extensions.dart
```diff
@@ -0,0 +1,519 @@
+// Copyright (c) 2020, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// Extension methods on iterables.
```
natebosch, line 5: This isn't a useful comment.

lib/src/iterable_extensions.dart
```diff
@@ -0,0 +1,519 @@
...
+
+import 'package:collection/src/utils.dart';
+
+import "algorithms.dart";
+
+/// Extensions that apply to all iterables.
```
natebosch, line 11: If the only thing we can say is to repeat the signature, we should omit the comment. Same thing for all the docs on the extensions in this file.

lib/src/iterable_extensions.dart
```diff
@@ -0,0 +1,519 @@
...
+extension IterableExtension<T> on Iterable<T> {
+
+  /// The elements that do not satisfy [test].
+  Iterable<T> whereNot(bool test(T element)) =>
+      where((element) => !test(element));
+
+  /// Creates a sorted list of the elements of the iterable.
+  ///
+  /// The elements are ordered by the [compare] [Comparator].
+  List<T> sorted(int compare(T a, T b)) => [...this]..sort(compare);
```
natebosch, line 21: Should we match the signature on `List.sort` and make the argument optional?

lib/src/iterable_extensions.dart
```diff
@@ -0,0 +1,519 @@
...
+    while (iterator.moveNext()) {
+      var element = iterator.current;
+      if (compare(previousElement, element) > 0) return false;
+      previousElement = element;
+    }
+    return true;
+  }
+
+  /// Whether the elements are sorted by the [compare] ordering of [keyOf].
+  bool isSortedBy<K>(K keyOf(T element), int compare(K a, K b)) {
```
natebosch, line 59: This is inconsistent with `sortedBy` and `sortedCompareBy`. Should we rename this `sortedCompareBy` and add a `sortedBy<K extends Comparable<K>>`?

lib/src/iterable_extensions.dart
```diff
@@ -0,0 +1,519 @@
...
+    var previousKey = keyOf(iterator.current);
+    while (iterator.moveNext()) {
+      var key = keyOf(iterator.current);
+      if (compare(previousKey, key) > 0) return false;
+      previousKey = key;
+    }
+    return true;
+  }
+
+  /// The elements of this iterable with duplicates removed.
```
natebosch, line 71: Does this offer enough value over `.toSet()` to make it worth adding?

lib/src/iterable_extensions.dart
```diff
@@ -0,0 +1,519 @@
...
+  Iterable<T> unique() sync* {
+    var seen = <T>{};
```
natebosch, line 73: Would `HashSet<T>();` be cheaper? We don't care about maintaining insertion order.

lib/src/iterable_extensions.dart
```diff
@@ -0,0 +1,519 @@
...
+    }
+    return result;
+  }
+
+  /// Groups elements into sets by [keyOf].
+  Map<K, Set<T>> groupSetsBy<K>(K keyOf(T element)) {
+    var result = <K, Set<T>>{};
+    for (var element in this) {
+      var key = keyOf(element);
+      result[key] = (result[key] ?? {})..add(element);
```
natebosch, line 207: I'm not sure if it actually makes a difference (or if we should be trying to micro-optimize without benchmarks) but I _think_ this is faster on dart2js with `??=`

```suggestion
      (result[key] ??= {}).add(element);
```

lib/src/iterable_extensions.dart
```diff
@@ -0,0 +1,519 @@
...
+  /// are emitted as a list.
+  /// Any final elements are emitted at the end.
+  ///
+  /// Example:
+  /// Example:
+  /// ```dart
+  /// var parts = [1, 2, 3, 4, 5, 6, 7, 8, 9].split(isPrime);
+  /// print(parts); // ([1], [2], [3, 4], [5, 6], [7, 8, 9])
+  /// ```
+  Iterable<List<T>> splitBefore(bool test(T element)) =>
```
natebosch, line 236: Is this speculative, or has someone asked for this?

lib/src/iterable_extensions.dart
```diff
@@ -0,0 +1,519 @@
...
+  /// The sum is zero if the iterable is empty.
+  T sum() {
+    var result = (0.5 is T ? 0.0 : 0) as T;
+    for (var value in this) {
+      result += value;
+    }
+    return result;
+  }
+
+  /// The average of the elements.
```
natebosch, line 383: We need to call out that this is the mean, not the median. Or we could name it `mean` or `arithmeticMean` and avoid the ambiguous `average` term altogether.

lib/src/iterable_extensions.dart
```diff
@@ -0,0 +1,519 @@
...
+    }
+  }
+}
+
+/// Extensions that apply to iterables of numbers.
+extension IterableNumberExtension<T extends num> on Iterable<T> {
+  /// The sum of the elements.
+  ///
+  /// The sum is zero if the iterable is empty.
+  T sum() {
```
natebosch, line 375: Should these be getters?

lib/src/iterable_extensions.dart
```diff
@@ -0,0 +1,519 @@
...
+    throw StateError("No element");
+  }
+
+  /// Creates a sorted list of the elements of the iterable.
+  ///
+  /// If the [compare] function is not supplied, the sorting uses the
+  /// natural [Comparable] ordering of the elements.
+  List<T> sorted([int compare(T a, T b)]) => [...this]..sort(compare);
+
+    /// Whether the elements are sorted by the [compare] ordering.
```
natebosch, line 483: The doc should describe the case where `compare` is omitted.

```dart
  /// Whether the element are sorted.
  ///
  /// The order is checked against [compare] if it is passed,
  /// otherwise it is checked against the element's natural order.
```

lib/src/list_extensions.dart
```diff
@@ -0,0 +1,281 @@
...
+  /// Calls [action] for each element along with the index in the
+  /// iteration order.
+  void forEachIndexed(void action(int index, E element)) {
+    for (var index = 0; index < length; index++) {
+      action(index, this[index]);
+    }
+  }
+
+  /// Maps each element and its index to a new value.
+  Iterable<R> mapIndexed<R>(R convert(int index, E element)) sync* {
```
natebosch, line 26: How much value are we buying by specializing this against `Iterable.mapIndexed`? Do we expect `this[index]` to be meaningfully different (faster?) than `for(var element in this)`?

lib/src/list_extensions.dart
```diff
@@ -0,0 +1,281 @@
...
+  /// Filters the elements on their value and index.
+  Iterable<E> whereIndexed(bool test(int index, E element)) sync* {
+    for (var index = 0; index < length; index++) {
+      var element = this[index];
+      if (test(index, element)) yield element;
+    }
+  }
+
+  /// Expands each element and index to a number of elements in a new iterable.
+  Iterable<R> expendIndexed<R>(Iterable<R> expend(int index, E element)) sync* {
```
natebosch, line 41: Typo

```suggestion
  Iterable<R> expandIndexed<R>(Iterable<R> expend(int index, E element)) sync* {
```

lib/src/list_extensions.dart
```diff
@@ -0,0 +1,281 @@
...
+    var index = 0;
+    for (var element in this) {
+      yield* expend(index++, element);
+    }
+  }
+
+
+  /// Sort a range of elements by [compare].
```
natebosch, line 49: I don't think we should add both this _and_ the ability to sort a `ListSlice`

lib/src/list_extensions.dart
```diff
@@ -0,0 +1,281 @@
...
+    end = RangeError.checkValidRange(start, end, length);
+    var self = this;
+    if (self is ListSlice) return self.slice(start, end);
+    return ListSlice<E>(this, start, end);
+  }
+
+  /// Whether [other] has the same elements as this list.
+  ///
+  /// Returns true iff [other] has the same [length]
+  /// as this list, and the elemets of this list and [other]
```
natebosch, line 110: Typo
```suggestion
  /// as this list, and the elements of this list and [other]
```

lib/src/list_extensions.dart
```diff
@@ -0,0 +1,281 @@
...
+  /// at the same indices are equal (according to `==`).
```
natebosch, line 111: Comment needs to be rephrased to account for the `equality` argument.

natebosch (changes requested):

lib/src/iterable_extensions.dart
lrhn: Seems reasonable.
And it should take start/end offsets too.

lib/src/iterable_extensions.dart
lrhn: Probably not, no.

lib/src/list_extensions.dart
lrhn: Yes, but maybe not by a lot.

lib/src/iterable_extensions.dart
lrhn: The signature of `List.sort` is actually bad. It allows you to omit the `compare`, but that only really works when the elements are `Comparable`. With extension methods we can do *better*.
There is a `sorted([int compare(E a, E b)])` on `Iterable<T extends Comparable<T>>`.

analysis_options.yaml
lrhn: Done for now.

lib/src/algorithms.dart
lrhn: There are multiple layers of API here. Both a general low-level API that people can use to build other libraries from, and which isn't necessarily optimized for ease-of-use, and a high-level convenience API that accesses the low-level API in various convenient manners. Actually, most of the former until now. It's a package of building blocks more than a package of solutions.

We could stop exposing the low-level API directly. That would mean not exporting the new quickSort or xBy methods from `algorithms` and only expose them as extension methods. It'll probably work. There will likely be someone asking us to expose it anyway, but if the full power is still accessible as an extension method, it's probably not an issue.

As for `binarySearch`, the reasons I originally didn't make it as an extension is that it doesn't actually work on all lists. It only works on *sorted* lists, and therefore I didn't want it to be an instance method on all lists. (For the same reason I think `sort` is badly designed as well, but that can be saved by extension methods, binary search cannot).
It's ... probably overly cautious of me, and I'll add the binary search as extension methods as well.

natebosch: > There will likely be someone asking us to expose it anyway, but if the full power is still accessible as an extension method, it's probably not an issue.

Waiting to expose something until someone asks is my preferred approach. As a team we've spent a non-trivial amount of time trying to remove stuff that has little or no usage.

mit-mit:
@lrhn looks stale?

lrhn:
PTAL

natebosch:
I had been waiting to review after the merge conflicts were resolved... Do you want me to do a first pass and then re-review after solving the conflicts?

lrhn:
Conflicts should be gone now.

lib/src/iterable_extensions.dart
```diff
@@ -0,0 +1,725 @@
...
+/// [IterableComparableExtension] and [IterableNullableExtension].
+extension IterableExtension<T> on Iterable<T> {
+  /// The elements that do not satisfy [test].
+  Iterable<T> whereNot(bool Function(T element) test) =>
+      where((element) => !test(element));
+
+  /// Creates a sorted list of the elements of the iterable.
+  ///
+  /// The elements are ordered by the [compare] [Comparator].
+  List<T> sorted(Comparator<T> compare) => [...this]..sort(compare);
```
natebosch, line 26: Should we make this an optional argument to match `List.sort`?

lib/src/iterable_extensions.dart
```diff
@@ -0,0 +1,725 @@
...
+      K Function(T element) keyOf, Comparator<K> compare) {
+    var elements = [...this];
+    quickSortBy<T, K>(elements, keyOf, compare);
+    return elements;
+  }
+
+  /// Whether the elements are sorted by the [compare] ordering.
+  bool isSorted(Comparator<T> compare) {
+    var iterator = this.iterator;
+    if (!iterator.moveNext()) return true;
```
natebosch, line 52: [nit] I think it would be worth calling out the behavior for an empty list in the doc comment.

lib/src/iterable_extensions.dart
```diff
@@ -0,0 +1,725 @@
...
+  /// The last element, or `null` if the iterable is empty.
+  T? get lastOrNull {
+    if (isEmpty) return null;
+    return last;
+  }
+
+  /// The single element satisfying [test].
+  ///
+  /// Returns `null` if there are either no elements
+  /// or more than one element satisfying [test].
```
natebosch, line 259: In the SDK I think the `singleWhere` methods throw if there is more than one match, and use `orElse` only for the no-match case. This is a different behavior that may be a better design, but also could be confusing.

If we do want this behavior, can we call out the difference specifically?

lib/src/iterable_extensions.dart
```diff
@@ -0,0 +1,725 @@
...
+    }
+    return result;
+  }
+
+  /// The artihmetic mean of the elements of a non-empty list.
+  ///
+  /// The arithmetic mean is the sum of the elements
+  /// divided by the number of elements.
+  ///
+  /// The list must not be empty.
```
natebosch, line 543: [nit] Replace all "list" in the comment with "iterable"

lib/src/iterable_extensions.dart
```diff
@@ -0,0 +1,725 @@
...
+  double get average {
```
natebosch, line 544: Should we go with an algorithm that won't overflow?

```dart
extension on Iterable<num> {
  double get average {
    var average = 0.0;
    var count = 1;
    for (var element in this) {
      average += (element - average) / count;
      count++;
    }
    return average;
  }
}
```

lib/src/iterable_extensions.dart
```diff
@@ -0,0 +1,725 @@
...
+      previousElement = element;
+    }
+    return true;
+  }
+}
+
+/// Extensions on comparator functions.
+extension ComparatorExtension<T> on Comparator<T> {
+  /// The inverse ordering of this comparator.
+  int Function(T, T) get inverse => (T a, T b) => this(b, a);
```
natebosch, line 706: [nit] Why use the function type here and the name `Comparator` below?

lib/src/iterable_extensions.dart
natebosch: I forgot I had already asked this and we resolved the conversation. Sorry about that!

https://github.com/dart-lang/collection/pull/135#discussion_r425658615

natebosch (approved):
Overall I'm _slightly_ worried about adding so many of these speculatively since there is a high cost to remove them, but given that they are extension methods I think the overall cost to keeping them is low.

CHANGELOG.md
```diff
@@ -3,6 +3,20 @@
 * Remove the unusable setter `UnionSetController.set=`. This was mistakenly
   added to the public API but could never be called.
 
+* Added `quickSort` to `algorithms.dart`.
+
+* Add extra optional `Random` argument to `shuffle`.
+
+* Generalize some of the functions in `algorithms.dart` to
```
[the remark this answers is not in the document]
lrhn: Done.  Also removing quickSort reference above, since it's also not exported.

lrhn (commented):
Addressed comments.

lib/src/algorithms.dart
lrhn: We're now not exposing the new top-level members.

PR #136, opened 2020-06-04
fix the wrapper tests for nnbd

jakemac53:
Make the mocks take an actual instance and use mirrors to delegate calls to that instance, returning the real result instead of null.

Closes https://github.com/dart-lang/collection/issues/127


jakemac53:
cc @kevmoo

jakemac53:
> Can you expand the commit message a bit to describe the use of mirrors?

Done

PR #137, opened 2020-06-07
Correct DeepCollectionEquality documentation

creativecreatorormaybenot:
That cost me a few, yikes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 * Deprecate `mapMap`. The Map interface has a `map` call and map literals can
   use for-loop elements which supersede this method.
+* Correct the `DeepCollectionEquality` documentation.
 
 ## 1.14.12
 
--- a/lib/src/equality.dart
+++ b/lib/src/equality.dart
@@ -409,10 +409,10 @@ class MultiEquality<E> implements Equality<E> {
 /// Recognizes lists, sets, iterables and maps and compares their elements using
 /// deep equality as well.  
 /// **Note**: the maximum depth for these comparisons is 1! This means that the
-/// comparison will fail when a sub collection in the collections to be compared
+/// comparison will fail when a sub collection in the collection to be compared
 /// contains another collection. For example, if you want to compare a list
-/// containing a list containing another list, [DeepCollectionEquality] will not
-/// be able to deeply compare the nested lists.
+/// containing a list containing another list, [DeepCollectionEquality] will 
+/// not be able to deeply compare the nested lists.
 ///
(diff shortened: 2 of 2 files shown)


lrhn:
I actually think `DeepCollectionEquality` should work to arbitrary depths. If it doesn't work, that's a bug in the implementation, not the documentation.

creativecreatorormaybenot:
@lrhn Yes, I thought so too. However, I wanted to believe that the implementation was intentional 😶

natebosch:
I cannot reproduce the bug. Do you have sample code that demonstrates `DeepCollectionEquality` only working to a depth of 1?

creativecreatorormaybenot:
@natebosch Just look at the code? 😄

It will e.g. use a `ListEquality` for lists in the root collection (but it should obviously use another `DeepCollectionEquality` for recursion, which does not work atm).

Example:

```dart
final list = [
  [
     [1, 2],
  ].
];
final list2 = [
  [
     [1, 2],
  ].
];

const DeepCollectionEquality().equals(list1, list2); // false
```

creativecreatorormaybenot:
@natebosch You just need to take a look at the source code 😄

It will e.g. use a `ListEquality` for lists in the root collection (but it should obviously use another `DeepCollectionEquality` for recursion, which does not work atm).

Example:

```dart
final list = [
  [
     [1, 2],
  ].
];
final list2 = [
  [
     [1, 2],
  ].
];

const DeepCollectionEquality().equals(list1, list2); // false
```

creativecreatorormaybenot:
@natebosch You just need to take a look at the source code 😄

It will e.g. use a `ListEquality` for lists in the root collection (but it should obviously use another `DeepCollectionEquality` for recursion, which does not work atm).

Example:

```dart
final list = [
  [
     [1, 2],
  ],
];
final list2 = [
  [
     [1, 2],
  ],
];

const DeepCollectionEquality().equals(list1, list2); // false
```

natebosch:
You example works for me on the master branch. Which version of the package are you using?

If you concern is this line of code:

https://github.com/dart-lang/collection/blob/80f5b6de8a8d8d584732a71bb59912da3e44883b/lib/src/equality.dart#L449

This is the correct implementation for arbitrary depth. Specifically the `this` argument to `ListEquality(this)` ensures that each element is compared also with `DeepCollectionEquality`.

Closing for now since this looks like a misunderstanding.

creativecreatorormaybenot:
@natebosch Thanks 👍🏽

I will need to figure out why it did not work for me. This can stay closed until then.

PR #142, opened 2020-06-19
Prep for merge back into master and publishing of the nnbd release

jakemac53:
Updates travis to run on be/raw/latest, and test with null safety enabled.

- Runs the analyzer on the `lib` dir without the explicit flag to verify the allow list functionality
- Runs the analyzer on the `test` dir with the flag since the allow list appears to have only been implemented for `package:` uris. cc @leafpetersen was this the intention? It seems wrong to me, given that the notion of a package has been expanded to its entire root directory.
- Runs tests on the vm and ddc, all in strong mode currently
- Updates the sdk constraint to `>=2.9.0-18.0 <2.9.0` (this allows dev or beta releases)
- Updates the release version to `1.15.0-nnbd`. Note that this is the actual version number I propose we publish as, cc @kevmoo .

If we want a solution for running weak mode (non-opted in) tests, that will need to happen separately.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,10 +3,37 @@ language: dart
 dart:
 - be/raw/latest
 
-dart_task:
-- test
-- dartfmt
-- dartanalyzer: --fatal-infos --fatal-warnings .
+jobs:
+  include:
+    - stage: analyze_and_format
+      name: "Analyze lib/"
+      dart: be/raw/latest
+      os: linux
+      script: dartanalyzer --fatal-warnings --fatal-infos lib/
+    - stage: analyze_and_format
+      name: "Analyze test/"
+      dart: be/raw/latest
+      os: linux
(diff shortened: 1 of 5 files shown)


natebosch:
> * Updates the release version to `1.15.0-nnbd`. Note that this is the actual version number I propose we publish as

I thought we were leaning more towards breaking changes. What is the current thinking on this?

jakemac53:
> I thought we were leaning more towards breaking changes. What is the current thinking on this?

At least for now we are planning on doing pre-release versions that are non-breaking while a decision is ultimately reached.

I added a line to the changelog that this may not represent the ultimate stable version for the nnbd release.

jakemac53:
At least for packages like `collection` though it does sound like we will end up releasing it as non-breaking. We might have more leeway when it comes to lesser-used packages.

PR #147, opened 2020-07-16
prep the release for 2.10 dev sdk

jakemac53:
Not quite ready to merge (we should publish the version for the 2.9 stable sdk first I think, which is why I let an extra version in there).

@whesse built us a custom sdk with the right version so that travis will be happy. It is a bit of a weird sdk in that still only requires language version `2.9` for our transitive deps, but that makes things a bit easier.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,27 +7,27 @@ jobs:
   include:
     - stage: analyze_and_format
       name: "Analyze lib/"
-      dart: preview/raw/2.10.0-0.0-dev
+      dart: preview/raw/2.10.0-0.1-dev
       os: linux
       script: dartanalyzer --fatal-warnings --fatal-infos lib/
     - stage: analyze_and_format
       name: "Analyze test/"
-      dart: preview/raw/2.10.0-0.0-dev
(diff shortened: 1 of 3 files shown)


natebosch:
Do we have any doc which lays out the plan for the different SDK and package versions? Are we going to have what that is 2.9-dev only, one 2.9-stable only and one 2.10-dev only?

jakemac53:
The plan is still evolving - I brought it up in an email thread with engprod when the stable branch was cut because there are some obvious issues.

The current plan is to publish a version which allows 2.9 dev and stable (just increase the upper bound) and cherry pick into flutter a commit that pins those versions. Then when they do flip to stable those packages will still work. It should be safe since we have already cut the stable branch so we don't expect any breakage, although breaking cherry picks could happen they would break the flutter bots so we would at least know.

jakemac53:
I am not sure why travis is unhappy here - locally the pub get runs fine :/.

I thought the overrides would allow the sdk constraints to not match up?

natebosch:
SGTM

Issue #140, opened 2020-06-18
HeapPriorityQueue doesn't let one iterate in unsorted order


sortie:
My code has several huge priority queues using `HeapPriorityQueue`, where it occasionally needs to consider all the elements in the queue where the order doesn't matter (e.g. to tally up how long it will take to process the queue, or to consider all the items and move a few to another queue if circumstances changes).

However, `HeapPriorityQueue` only offers `toList()` and `toSet()` for this purpose. However, they spend O(nlogn) to sort the queue (and I need a linear time algorithm) and duplicates the data structure (I just need constant memory). The only way to iterate the queue in linear time in unsorted order is to use `removeAll()`, however it is destructive.

The `HeapPriorityQueue` internally just uses a list to manage the data structure, which is what is returned by  `removeAll()`. It could easily gain a `Iterable<E> get unsorted` getter or something that returns the backing data structure with the appropriate concurrent modification protections.

This is probably a breaking change and needs a major version bump in case anyone implemented `HeapPriorityQueue`.

I'm currently working around the issue associating each priority queue with a `Set` in my mod, however it duplicates the amount of memory spent on the queues.

lrhn:
Adding a member to an implementation class is not a breaking change, so we can add it to `HeapPriorityQueue`. It's potentially breaking to add it to the interface `PriorityQueue`, since that is intended for general use, and someone might be implementing it.

It's somewhat annoying to add an operation which is not general to all `PriorityQueue`s, but since some priority queues could easily keep objects in priority order all the time, an `unorderedElements` iterable getter wouldn't make sense for them. 
The class interface itself does not implement `Iterable`, so we haven't promised any way to access elements other than the "highest priority" one, and some priority queues might not even be able to do so (unlikely, but possible).
So, adding just to `HeapPriorityQueue` can be reasonable.

If we do it, I'd suggest `Iterable<E> get unorderedElements;`, documented to say that you must not modify the queue during iteration, and that you must not modify the elements to change their priority (in that case, do a `removeAll` followed by a rebuild).

lrhn:
I've added `toUnorderedList` to `PriorityQueue`. Let's see if anyone implements it.

sortie:
The `toUnorderedList` implementation solves the runtime complexity issue by reducing it from O(nlogn) to O(n). However, it still uses O(n) time and memory to duplicate the queue. It looks like this feature got bundled with null safety so my service can't use it until null safety is released.

[...]

PR #152, opened 2020-09-10
Add `toUnorderedList` to `PriorityQueue`.

### LINKED ISSUE #45: PriorityQueue and comparison function inconsistent
The documentat of `PriorityQueue` states:
```
  /// The [comparison] is a [Comparator] used to compare the priority of
  /// elements. An element that compares as less than another element has
  /// a higher priority.
```

(btw: the documentation doesn't state what the function should return).

However, the comparison function is also used as equality function.
For example, `PriorityQueue.contains`, or `PriorityQueue.remove` both use this function to determine equality. In some cases that works, but a common use case is to extract a priority field from one of many fields.

Example:
``` dart
class PriorityPoint {
  int x, y, priority;
  PriorityPoint(this.x, this.y, this.priority);
}

var queue = new PriorityQueue(comparison: (p1, p2) => p2.priority - p1.priority);
queue.add(new PriorityPoint(0, 1, 5));
queue.contains(new PriorityPoint(-1, -3, 5));  // Should probably return false.
```

I'm not sure what the best solution is, but the current behavior and documentation are not consistent.
Furthermore, the current behavior makes it hard to implement the provided use case.

lrhn:
Also change `HeapPriorityQueue` to use `==` for `contains` and `remove` checks.

Fix a bad null-safety migration for HeapPriorityQueue, and add tests to ensure it works with `null` values.

Fixes #45, #140

### DIFF
--- a/lib/src/priority_queue.dart
+++ b/lib/src/priority_queue.dart
@@ -10,7 +10,7 @@ import 'utils.dart';
 ///
 /// The queue allows adding elements, and removing them again in priority order.
 /// The same object can be added to the queue more than once.
-/// There is no specified oredering for objects with the same priority
+/// There is no specified ordering for objects with the same priority
 /// (where the `comparison` function returns zero).
 ///
 /// Operations which care about object equality, [contains] and [remove],
--- a/lib/src/priority_queue.dart
+++ b/lib/src/priority_queue.dart
@@ -9,6 +9,18 @@ import 'utils.dart';
 /// A priority queue is a priority based work-list of elements.
 ///
 /// The queue allows adding elements, and removing them again in priority order.
+/// The same object can be added to the queue more than once.
+/// There is no specified oredering for objects with the same priority
+/// (where the `comparison` function returns zero).
+///
+/// Operations which care about object equality, [contains] and [remove],
+/// use [Object.==] for testing equality.
+/// In most situations this will be the same as identity ([identical]),
+/// but there are types, like [String], where users can reasonably expect
+/// distinct objects to represent the same value.
+/// If elements override [Object.==], the `comparison` function must
+/// always give equal objects the same priority,
+/// otherwise [contains] or [remove] might not work correctly.
 abstract class PriorityQueue<E> {
   /// Creates an empty [PriorityQueue].
   ///
@@ -36,6 +48,14 @@ abstract class PriorityQueue<E> {
   /// Checks if [object] is in the queue.
   ///
   /// Returns true if the element is found.
+  ///
+  /// Uses the [Object.==] of elements in the queue to check
+  /// for whether they are equal to [object].
+  /// Equal objects objects must have the same priority
+  /// according to the [comparison] function.
+  /// That is, if `a == b` then `comparison(a, b) == 0`.
+  /// If that is not the case, this check might fail to find
+  /// an object.
   bool contains(E object);
 
   /// Adds element to the queue.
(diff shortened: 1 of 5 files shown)


jakemac53 (commented):
Is it actually useful to store nulls in a queue like this? That seems really dubious to me. We could instead make it `<E extends Object>`?

Maybe not worth restricting though...

lib/src/priority_queue.dart
```diff
@@ -9,6 +9,18 @@ import 'utils.dart';
 /// A priority queue is a priority based work-list of elements.
 ///
 /// The queue allows adding elements, and removing them again in priority order.
+/// The same object can be added to the queue more than once.
+/// There is no specified oredering for objects with the same priority
```
natebosch, line 13:
```suggestion
/// There is no specified ordering for objects with the same priority
```

lib/src/priority_queue.dart
```diff
@@ -223,18 +264,23 @@ class HeapPriorityQueue<E> implements PriorityQueue<E> {
...
     var set = SplayTreeSet<E>(comparison);
     for (var i = 0; i < _length; i++) {
-      set.add(_queue[i]!);
+      set.add(_elementAt(i));
     }
     return set;
   }
 
+  @override
+  List<E> toUnorderedList() => _toUnorderedList();
```
natebosch, line 279: Why do we need the private one? Can we put the implementation here and skip the function?

natebosch (approved):

lib/src/priority_queue.dart
lrhn: I call this method from `toList` too. It's an implementation detail that the two methods share code, so I don't want to risk someone overriding `toUnorderedList` and also affecting `toList`.
Generally, I avoid calling an overridable method from another method unless the latter method is explicitly documented as depending on the former. Code sharing is an implementation detail, so it shouldn't affect the visible API.

lrhn:
All our other collections accept `null`, so it would stand out if this one doesn't, and it would be a breaking change.
Using `as E` also works better than `!` with legacy code which might store `null`  values in a `HeapPriorityQueue<int>`, because `null as int*` doesn't throw, but `null!` does.

There is no real reason to not to accept `null` as a value. The `null` value is an object too.

The `_elementAt` code will only read a `null` if the `E` type actually accepts `null` - it's only used for indices in the range where we know we have stored elements. All the `as E` operations will only see `null` values actually stored into the heap, so it's safely compilable to JavaScript where we omit the `as` casts.

natebosch (approved):

jakemac53 (approved):

mit-mit:
@lrhn ready to merge?

lib/src/priority_queue.dart
```diff
@@ -267,10 +316,10 @@ class HeapPriorityQueue<E> implements PriorityQueue<E> {
     // in the heap will also have lower priority.
     do {
       var index = position - 1;
-      var element = _queue[index]!;
+      var element = _elementAt(index);
       var comp = comparison(element, object);
-      if (comp == 0) return index;
-      if (comp < 0) {
+      if (comp <= 0) {
+        if (comp == 0 && element == object) return index;
```
natebosch, line 322: This turned out to break at least one project's tests.

I think it's the right thing to do, and they were able to work around it.

lrhn: Super. If you rely on `compareTo` for equality, and don't override `==` to match, then I guess this would affect you.

PR #155, opened 2020-09-23
Add `unorderedElements` getter to PriorityQueue.

lrhn:
Also fix bug in `queue_list.dart` which made its tests fail.

### DIFF
--- a/lib/src/priority_queue.dart
+++ b/lib/src/priority_queue.dart
@@ -214,10 +214,12 @@ class HeapPriorityQueue<E> implements PriorityQueue<E> {
 
   @override
   void addAll(Iterable<E> elements) {
-    _modificationCount++;
+    var modified = 0;
     for (var element in elements) {
+      modified = 1;
       _add(element);
     }
+    _modificationCount += modified;
   }
 
   @override
@@ -466,16 +468,16 @@ class _UnorderedElementsIterable<E> extends Iterable<E> {
 
 class _UnorderedElementsIterator<E> implements Iterator<E> {
   final HeapPriorityQueue<E> _queue;
-  final int _modificationCount;
+  final int _initialModificationCount;
   E? _current;
   int _index = -1;
 
   _UnorderedElementsIterator(this._queue)
-      : _modificationCount = _queue._modificationCount;
+      : _initialModificationCount = _queue._modificationCount;
 
   @override
   bool moveNext() {
-    if (_modificationCount != _queue._modificationCount) {
+    if (_initialModificationCount != _queue._modificationCount) {
       throw ConcurrentModificationError(_queue);
     }
     var nextIndex = _index + 1;
--- a/lib/src/priority_queue.dart
+++ b/lib/src/priority_queue.dart
@@ -58,6 +58,13 @@ abstract class PriorityQueue<E> {
   /// an object.
   bool contains(E object);
 
+  /// Provides access to all the elements curently in the queue.
+  ///
+  /// The elements are iterated in no particular order.
+  /// The order is stable as long as the queue is not modified.
+  /// The queue must not be modified during an iteration.
+  Iterable<E> get unorderedElements;
+
   /// Adds element to the queue.
   ///
   /// The element will become the next to be removed by [removeFirst]
@@ -177,6 +184,11 @@ class HeapPriorityQueue<E> implements PriorityQueue<E> {
   /// The heap is implemented in the first [_length] entries of [_queue].
   int _length = 0;
 
+  /// Modification count.
+  ///
+  /// Used to detect concurrent modifications during iteration.
+  int _modificationCount = 0;
+
   /// Create a new priority queue.
   ///
   /// The [comparison] is a [Comparator] used to compare the priority of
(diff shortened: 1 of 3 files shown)


lib/src/priority_queue.dart
```diff
@@ -58,6 +58,13 @@ abstract class PriorityQueue<E> {
   /// an object.
   bool contains(E object);
 
+  /// Provides access to all the elements curently in the queue.
```
sortie, line 61: Mention the 'no particular order' in the leading paragraph as well since that's the essential purpose.

lib/src/priority_queue.dart
```diff
@@ -418,3 +445,41 @@ class HeapPriorityQueue<E> implements PriorityQueue<E> {
     _queue = newQueue;
   }
 }
+
+/// Implementation of [HeapPriorityQueue.unorderedElements].
+class _UnorderedElementsIterable<E> extends Iterable<E> {
+  final HeapPriorityQueue<E> _queue;
+  _UnorderedElementsIterable(this._queue);
+  @override
+  Iterator<E> get iterator => _UnorderedElementsIterator<E>(_queue);
```
sortie, line 454: Can we simply use _queue._queue.iterator and then use the concurrent modification checks already done by the List class? I guess not, because of removeAll() and maybe other operations (like adding a duplicate element) won't properly catch concurrent modifications.

sortie (approved):
This nicely solves my needs.

Should any CHANGELOG or such be updated?

lib/src/priority_queue.dart
lrhn: That's actually not the essential purpose, the essential purpose is to be efficient.
Being arbitrarily ordered is a means to that goal. If it didn't help us being efficient, we wouldn't care that the order is unspecified.

lib/src/priority_queue.dart
lrhn: We can't. The only thing `list.iterator` checks for is length changes. A list is treated like a map from integers to values, and the only structural change is changing which integers are allowed - the length. Changing values is like changing the values of a `Map`, it's not something which affects iteration. (Corollary: a fixed length list doesn't need to check for concurrent modifications).

Also, the `_queue` backing store is a `List<E?>` which may be longer than the number of elements. We need to restrict it to the first `length` elements and cast it to `E`. 
A `_queue.take(length).cast<E>()` would do it, but less efficiently, and wouldn't catch most modifications.

The modifications to worry about are those which causes an iteration to repeat or skip an element which was in the queue during the entire iteration. That's what I consider an *inconsistent* iteration. 
Not emitting an element which was removed or added during iteration is not a problem. Emitting an element after it was removed is not great, but acceptable - it was part of the collection that was originally asked to be iterated.
For a queue like this, most operations change the position of elements, which means that they risk duplicating or skipping elements when we are iterating by index. A `removeAll` is not actually a problem, we could allow iteration to continue (and then immediately stop because there is no next element), but operations adding or removing single elements are.

I originally wrote
```dart
Iterable<E> get unorderedElements sync* {
  var modificationCount = _modificationCount;
  for (var i = 0; i < _length; i++) {
    yield _queue[i] ?? null as E;
    if (modificationCount != _modificationCount) {
      throw ConcurrentModificationError(this);
    }
  }
}
```
but I'm assuming that `sync*` implementations will probably also be less efficient than a hand-written implementation (in both code size and performance).

lib/src/priority_queue.dart
lrhn: I've update the documentation.

lib/src/priority_queue.dart
sortie: Oh wow I didn't need that thorough an answer but that was interesting to read.

lib/src/priority_queue.dart
```diff
@@ -418,3 +455,41 @@ class HeapPriorityQueue<E> implements PriorityQueue<E> {
...
+}
+
+class _UnorderedElementsIterator<E> implements Iterator<E> {
+  final HeapPriorityQueue<E> _queue;
+  final int _modificationCount;
```
jakemac53, line 469: nit: `_initialModificationCount` or similar to indicate this is the count when this was created.

test/priority_queue_test.dart
```diff
@@ -69,6 +69,11 @@ void testQueueBody<T>(
 
     expect(q.toList(), equals(elements));
     expect(q.toSet().toList(), equals(elements));
+    expect(q.toUnorderedList(), unorderedEquals(elements));
+    expect(q.unorderedElements, unorderedEquals(elements));
```
jakemac53, line 73: Given the custom logic for concurrent modification it would be good to test that

lib/src/priority_queue.dart
lrhn: I don't like to just say "no" like Terkel's dad :grin:.

test/priority_queue_test.dart
lrhn: Done.

lib/src/priority_queue.dart
lrhn: Done.

PR #157, opened 2020-09-28
General clean-up and tweaks.

lrhn:
Fixes some null-safety issues, usees some newer features (const classes can use mixins now),
and replaces unnecessary `late` with a dual constructor.

Fixes the `CombinedList.iterator` having potentially quadratic complexity
since the default list iterator does repeated indexing.

Ensure that interfaces returning an `Iterable` does not return a `Set` instead.

### DIFF
--- a/lib/src/combined_wrappers/combined_iterator.dart
+++ b/lib/src/combined_wrappers/combined_iterator.dart
@@ -1,0 +1,40 @@
+// Copyright (c) 2020, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// The iterator for `CombinedIterableView` and `CombinedListView`.
+///
+/// Moves through each iterable's iterator in sequence.
+class CombinedIterator<T> implements Iterator<T> {
+  /// The iterators that this combines, or `null` if done iterating.
+  ///
+  /// Because this comes from a call to [Iterable.map], it's lazy and will
+  /// avoid instantiating unnecessary iterators.
+  Iterator<Iterator<T>>? _iterators;
+
+  CombinedIterator(Iterator<Iterator<T>> iterators) : _iterators = iterators {
+    if (!iterators.moveNext()) _iterators = null;
+  }
+
+  @override
+  T get current {
+    var iterators = _iterators;
+    if (iterators != null) return iterators.current.current;
+    return null as T;
+  }
+
+  @override
+  bool moveNext() {
+    var iterators = _iterators;
+    if (iterators != null) {
+      do {
+        if (iterators.current.moveNext()) {
+          return true;
+        }
+      }
+      while (iterators.moveNext());
+      _iterators = null;
+    }
+    return false;
+  }
+}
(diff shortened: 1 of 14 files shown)


lib/src/combined_wrappers/combined_iterator.dart
```diff
@@ -0,0 +1,40 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// The iterator for `CombinedIterableView` and `CombinedListView`.
+///
+/// Moves through each iterable's iterator in sequence.
+class CombinedIterator<T> implements Iterator<T> {
+  /// The iterators that this combines, or `null` if done iterating.
+  ///
+  /// Because this comes from a call to [Iterable.map], it's lazy and will
```
natebosch, line 11: I worry this comment will become stale now that it is no longer an implementation detail of a single library. Since it should still be private to the package this might not be a huge deal, but we should consider moving the comment to the constructor to be something like "iterators should not eagerly instantiate iterators which have not been reached".

lib/src/empty_unmodifiable_set.dart
```diff
@@ -4,65 +4,44 @@
...
 
+import 'package:collection/collection.dart';
+
 import 'unmodifiable_wrappers.dart';
 
-// Unfortunately, we can't use UnmodifiableSetMixin here, since const classes
-// can't use mixins.
-/// An unmodifiable, empty set that can have a const constructor.
+/// An unmodifiable, empty set which can be constant.
 class EmptyUnmodifiableSet<E> extends IterableBase<E>
```
natebosch, line 12: [out of scope for this PR] Is this class buying value over `const <E>{}`? Should we consider deprecating it?

lib/src/union_set_controller.dart
```diff
@@ -23,20 +23,21 @@ import 'union_set.dart';
 /// ```
 class UnionSetController<E> {
   /// The [UnionSet] that provides a view of the union of sets in [this].
-  UnionSet<E> get set => _set;
-  late final UnionSet<E> _set;
+  final UnionSet<E> set;
 
   /// The sets whose union is exposed through [set].
-  final _sets = <Set<E>>{};
+  final _sets;
```
natebosch, line 29: Can we add a type?
```suggestion
  final Set<Set<E>> _sets;
```

natebosch (approved):

lib/src/combined_wrappers/combined_iterator.dart
lrhn: It is still an implementation detail, the class is not exported by the package.

lib/src/empty_unmodifiable_set.dart
lrhn: Probably not worth it any more. It's a valid superclass, but I can't see for what.
Yes, we should consider deprecating it.

lib/src/union_set_controller.dart
lrhn: Doh,  yes. That's why I prefer always writing the types, they keep working when you remove the initializer!

PR #158, opened 2020-09-30
Add `choose` extension method to iterable, choosing a number of elements at random.

lrhn:
Inspired by: https://stackoverflow.com/questions/64117939/picking-n-unique-random-enums-in-dart

### DIFF
--- a/lib/src/iterable_extensions.dart
+++ b/lib/src/iterable_extensions.dart
@@ -2,6 +2,8 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'dart:math' show Random;
+
 import 'package:collection/src/utils.dart';
 
 import 'algorithms.dart';
@@ -16,6 +18,37 @@ import 'algorithms.dart';
 /// iterables with specific element types include those of
 /// [IterableComparableExtension] and [IterableNullableExtension].
 extension IterableExtension<T> on Iterable<T> {
+  /// Chooses [count] elements at random from this iterable.
+  ///
+  /// There must be at least [count] elements in the iterable.
+  /// The returned list contains [count] different elements
+  /// of the iterable.
+  /// If the same value occurs more than once in the iterable,
+  /// it can also occur more than once in the chosen elements.
+  ///
+  /// Each element of the iterable has the same chance of
+  /// being chosen.
+  /// The chosen elements are not in any specific order.
+  List<T> choose(int count, [Random? random]) {
+    RangeError.checkNotNegative(count, 'count');
+    var iterator = this.iterator;
+    var chosen = [
+      for (var i = 0; i < count; i++)
+        if (iterator.moveNext())
+          iterator.current
+        else
+          throw StateError('Too few elements')
+    ];
+    var index = count;
+    random ??= Random();
+    while (iterator.moveNext()) {
+      index++;
+      var position = random.nextInt(index);
+      if (position < count) chosen[position] = iterator.current;
+    }
+    return chosen;
+  }
+
   /// The elements that do not satisfy [test].
   Iterable<T> whereNot(bool Function(T element) test) =>
       where((element) => !test(element));
@@ -740,7 +773,7 @@ extension IterableComparableExtension<T extends Comparable<T>> on Iterable<T> {
 /// Extensions on comparator functions.
 extension ComparatorExtension<T> on Comparator<T> {
   /// The inverse ordering of this comparator.
-  int Function(T, T) get inverse => (T a, T b) => this(b, a);
+  Comparator<T> get inverse => (T a, T b) => this(b, a);
 
   /// Makes a comparator on [R] values using this comparator.
   ///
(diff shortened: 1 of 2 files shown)


test/extensions_test.dart
```diff
@@ -989,6 +989,52 @@ void main() {
...
+      test('single', () {
+        var single = iterable([1]);
+        expect(single.choose(0), []);
+        expect(single.choose(1), [1]);
+      });
+      test('multiple', () {
+        var multiple = iterable([1, 2, 3, 4, 5, 6, 7, 8, 9, 10]);
+        expect(multiple.choose(0), []);
+        var one = multiple.choose(1);
+        expect(one.length, 1);
```
natebosch, line 1011:
```suggestion
        expect(one, hasLength(1));
```

test/extensions_test.dart
```diff
@@ -989,6 +989,52 @@ void main() {
...
+        expect(one.first, inInclusiveRange(1, 10));
+        var seen = <int>{};
+        do {
+          var some = multiple.choose(3);
+          expect(some.length, 3);
```
natebosch, line 1016:
```suggestion
          expect(some, hasLength(3));
```

test/extensions_test.dart
```diff
@@ -989,6 +989,52 @@ void main() {
...
+          expect(some[0], inInclusiveRange(1, 10));
+          expect(some[1], inInclusiveRange(1, 10));
+          expect(some[2], inInclusiveRange(1, 10));
+          expect(some[0], isNot(some[1]));
+          expect(some[0], isNot(some[2]));
+          expect(some[1], isNot(some[2]));
+          seen.addAll(some);
+        } while (seen.length < 10);
+        // Should eventually terminate.
```
natebosch, line 1025: note that since there is a call to `expect` in the loop, if for some reason this _doesn't_ terminate it also won't time out in the test runner.

I don't think this is a real risk here, but the pattern does make me nervous if we applied it elsewhere.

natebosch (approved):
Should we consider adding a `Stream.choose` in `package:async`?

test/extensions_test.dart
lrhn: I *really* dislike the `test` package for needing to know a large number of idiosyncratic matchers by name.

test/extensions_test.dart
lrhn: It should probably make us nervous here too, I'll unroll the loop once so there are no expects inside the loop.

lib/src/iterable_extensions.dart
```diff
@@ -16,6 +18,37 @@ import 'algorithms.dart';
...
+  /// There must be at least [count] elements in the iterable.
+  /// The returned list contains [count] different elements
+  /// of the iterable.
+  /// If the same value occurs more than once in the iterable,
+  /// it can also occur more than once in the chosen elements.
+  ///
+  /// Each element of the iterable has the same chance of
+  /// being chosen.
+  /// The chosen elements are not in any specific order.
+  List<T> choose(int count, [Random? random]) {
```
lrhn, line 32: This variant throws if there are too few.

Should we also have one which just returns all the elements if there are fewer than `count`?
Say `chooseMax(int count, [Random? random])`.

Or make that the default behavior?

(I don't want a parameter to choose between the two behaviors in one function, that's unnecessarily messy, and I dislike using booleans to switch between different algorithms).

test/extensions_test.dart
natebosch: Yeah, we've had some discussions in the past about writing a nicer API using extension methods where potentially `hasLength` could be surfaced as a suggestion by the analyzer if it was an extension on `Expect<List>` (and we'd need to do the same for other common types with lengths). Then we could do things like `expect(some).hashLength(3)`. It wouldn't help if the first instinct is to write `expect(some.length).equals(3)`, but it might still help some with discoverability.

It's really unfortunate that the existing undiscoverable API is so baked in to `package:test` because it makes it hard to ship something better. I do have some ideas around loosening the coupling to make it feasible to ship alternative APIs, but I don't know if or when we'll be able to get them on OKRs.

lib/src/iterable_extensions.dart
natebosch: I also dislike the idea of a boolean argument.

I'm not excited about needing two methods for this, but I don't have a sense about which is more useful. The version that _doesn't_ throw can be used as a basis for the version that does throw, so it's more general, but also adds noise in the code if that isn't the behavior that the user was looking for.

I checked other languages/libraries and I couldn't find this method implemented in many of them. The python version throws, while the ruby version returns as many elements as it can (interestingly it also shuffles them).

While I was looking though, it did occur to me that most readers might find "sample" more intuitive than "choose". Even I associate "choose" more with the combinatorics operation that only considers the size, not the specific elements.

WDYT about renaming to "sample", and going with the version that does not throw for now. If we find usage is high enough, and enough folks are writing `if (sample.length < desiredLength) throw` we could add a second version.

lrhn:
Probably not that important to have the function on `Stream`. I think it's too rare a use-case.

PR #160, opened 2020-10-21
Fix typo in expandIndexed

natebosch:


jakemac53 (approved):
I assume this hasn't yet been released?

natebosch:
It hasn't been, but now it is.

My inclination is to just go with it. The only broken release is a `-nullsafety.4` and fixing this now should limit the exposure.

WDYT?

lrhn:
LGTM, thanks and sorry.
(And obviously it happened to a method which existed on `Iterable` with the same behavior, because otherwise it would have been very visible).

jakemac53:
Yes I am fine with just fixing this since we are still in pre-release

PR #161, opened 2020-10-22
Add `iterableEquals`, `listEquals`, etc. extension methods.

lrhn:
Also fixes type issue with `UnorderedIterableEquality`.
FYI @johnniwinther

### DIFF
--- a/lib/src/equality.dart
+++ b/lib/src/equality.dart
@@ -202,14 +202,14 @@ class ListEquality<E> implements Equality<List<E>> {
   bool isValidKey(Object? o) => o is List<E>;
 }
 
-abstract class _UnorderedEquality<E, T extends Iterable<E>?>
+abstract class _UnorderedEquality<E, T extends Iterable<E>>
     implements Equality<T> {
   final Equality<E> _elementEquality;
 
   const _UnorderedEquality(this._elementEquality);
 
   @override
-  bool equals(T elements1, T elements2) {
+  bool equals(T? elements1, T? elements2) {
     if (identical(elements1, elements2)) return true;
     if (elements1 == null || elements2 == null) return false;
     var counts = HashMap(
@@ -232,7 +232,7 @@ abstract class _UnorderedEquality<E, T extends Iterable<E>?>
   }
 
   @override
-  int hash(T elements) {
+  int hash(T? elements) {
     if (elements == null) return null.hashCode;
     var hash = 0;
     for (E element in elements) {
@@ -251,7 +251,7 @@ abstract class _UnorderedEquality<E, T extends Iterable<E>?>
 /// Two iterables are considered equal if they have the same number of elements,
 /// and the elements of one set can be paired with the elements
 /// of the other iterable, so that each pair are equal.
-class UnorderedIterableEquality<E> extends _UnorderedEquality<E, Iterable<E>?> {
+class UnorderedIterableEquality<E> extends _UnorderedEquality<E, Iterable<E>> {
   const UnorderedIterableEquality(
       [Equality<E> elementEquality = const DefaultEquality<Never>()])
       : super(elementEquality);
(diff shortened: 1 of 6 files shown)


lib/src/equality.dart
```diff
@@ -272,7 +272,7 @@ class UnorderedIterableEquality<E> extends _UnorderedEquality<E, Iterable<E>?> {
 /// The [equals] and [hash] methods accepts `null` values,
 /// even if the [isValidKey] returns `false` for `null`.
 /// The [hash] of `null` is `null.hashCode`.
-class SetEquality<E> extends _UnorderedEquality<E, Set<E>?> {
+class SetEquality<E> extends _UnorderedEquality<E, Set<E>> {
```
lrhn, line 275: This made `SetEquality<int>` implement `Equality<Set<int>?>`, which means that writing:
```dart
Equality<Set<X>> eq = SetEquality<X>(xEquality);
```
would be a type error. Seems nobody hit that yet, but fixing it to be certain.

(We still let `SetEquality<X>` accept `Set<X>?` as arguments to equality, but that's just being permissive in what it accepts, not in what it *is*).

natebosch: Could we handle this change in a separate PR?

lib/src/iterable_extensions.dart
```diff
@@ -18,6 +19,38 @@ import 'algorithms.dart';
...
 /// [IterableComparableExtension] and [IterableNullableExtension].
 extension IterableExtension<T> on Iterable<T> {
+  /// Whether this iterable has the same element order as [elements].
+  ///
+  /// Checks whether this iterable has the same elements as [elements],
+  /// in the same iteration order.
+  ///
+  /// Uses [equality] for determining if the values are the same.
+  /// If [equality] is omitted, the default `==` behavior is used.
+  bool iterableEquals(Iterable<T> elements, [Equality<T>? equality]) {
```
natebosch, line 29: This reads strangely to me. What do you think of `equalsIterable` instead?

lib/src/iterable_extensions.dart
```diff
@@ -18,6 +19,38 @@ import 'algorithms.dart';
 /// iterables with specific element types include those of
 /// [IterableComparableExtension] and [IterableNullableExtension].
 extension IterableExtension<T> on Iterable<T> {
+  /// Whether this iterable has the same element order as [elements].
```
natebosch, line 22: I think this should read "same values and same order as"

lib/src/iterable_extensions.dart
```diff
@@ -18,6 +19,38 @@ import 'algorithms.dart';
...
+  ///
+  /// Checks whether this iterable has the same elements as [elements],
```
natebosch, line 24: This feels like a duplicate sentence.

lib/src/iterable_extensions.dart
```diff
@@ -18,6 +19,38 @@ import 'algorithms.dart';
...
+  }
+
+  /// Whether this iterable has the same unordered elements as [elements].
+  ///
+  /// Checks whether this iterable has the same elements as [elements],
+  /// but not necessarily in the same iteration order.
+  ///
+  /// Uses [equality] for determining if the values are the same.
+  /// If [equality] is omitted, the default `==` behavior is used.
+  bool iterableUnorderedEquals(Iterable<T> elements, [Equality<T>? equality]) {
```
natebosch, line 48: How about `unorderedEqualsIterable`

lib/src/iterable_extensions.dart
```diff
@@ -793,3 +826,66 @@ extension ComparatorExtension<T> on Comparator<T> {
         return result;
       };
 }
+
+/// Extensions on sets.
```
natebosch, line 830: [nit] This comment can be omitted unless we have something more interesting to add.

lib/src/iterable_extensions.dart
```diff
@@ -793,3 +826,66 @@ extension ComparatorExtension<T> on Comparator<T> {
...
+  /// Whether this set has the same unordered elements as [elements].
+  ///
+  /// Checks whether this set has the same elements as [elements],
+  /// but not necessarily in the same iteration order.
+  ///
+  /// Uses [equality] for determining if the values are the same.
+  /// If [equality] is omitted, default to the equality used by
+  /// this set.
+  /// _(Notice: This differs from [SetEquality],
+  /// which is symmetric in the arguments)_.
```
natebosch, line 841: I don't think I understand this statement, and it doesn't look like the `SetEquality` docs mention "symmetric". Can you expand this?

mit-mit:
Closing as assumed stale

lrhn:
Doesn't seem fixed. The `SetEquality` type problem is still there.

PR #172, opened 2020-12-11
Add an empty unmodifiable map implementation

donny-dont:
Adds an `EmptyUnmodifiableMap` implementation which provides a `const` constructor.

The `Mixin` classes within `unmodifiable_wrappers.dart` were changed from `abstract class` to `mixin` to modernize the code. This allowed the `EmptyUnmodifiableMap` to have a `const` constructor without also defining a `const` constructor in `UnmodifiableMapMixin`.

### DIFF
--- a/lib/src/empty_unmodifiable_map.dart
+++ b/lib/src/empty_unmodifiable_map.dart
@@ -1,0 +1,40 @@
+// Copyright (c) 2020, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:collection';
+
+import 'unmodifiable_wrappers.dart';
+
+/// An unmodifiable, empty map which can be constant.
+class EmptyUnmodifiableMap<K, V>
+    with UnmodifiableMapMixin<K, V>
+    implements UnmodifiableMapView<K, V> {
+  const EmptyUnmodifiableMap();
+
+  @override
+  V? operator [](Object? key) => null;
+  @override
+  Map<RK, RV> cast<RK, RV>() => EmptyUnmodifiableMap<RK, RV>();
+  @override
+  bool containsKey(Object? key) => false;
+  @override
+  bool containsValue(Object? value) => false;
+  @override
+  Iterable<MapEntry<K, V>> get entries => Iterable<MapEntry<K, V>>.empty();
+  @override
+  void forEach(void Function(K key, V value) f) {}
+  @override
+  bool get isEmpty => true;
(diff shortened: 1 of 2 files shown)


donny-dont:
@kevmoo @natebosch I think this would help simplify some of the `shelf` stuff around context and headers.

```dart
class Context extends UnmodifiableMapView<String, Object> {
  factory Context(Map<String, Object> from) {
    if (from is Context) return from;
    if (from.isEmpty) return Context.empty();
    return Context._(from);
  }
  const factory Context.empty() = _EmptyContext;
  Context._(Map<String, Object> from) : super(from);
}

class _EmptyContext extends EmptyUnmodifiableMap<String, Object>
    implements Context {
  const _EmptyContext();
}
```

kevmoo:
Ooo! 

Tests?
Changelog entry?
Version bump?

donny-dont:
Sure! Wanted to make sure that the changes were acceptable before writing tests.

Version would be 1.15.0-nullsafety.6? I'm assuming after null safety lands the package is being bumped to 2.0 right?

kevmoo:
Let @natebosch chime in before you do any more...

natebosch:
Could we skip the extra class and use `const {}`?

```dart
class Context extends UnmodifiableMapView<String, Object> {
  factory Context(Map<String, Object> from) {
    if (from is Context) return from;
    if (from.isEmpty) return empty;
    return Context._(from);
  }
  static final empty = Context._(const {});
  Context._(Map<String, Object> from) : super(from);
}
```

Or, if we make the `UnmodifiableMap` constructor `const` it opens up options to get you closer to the pattern you had if you strongly prefer that:

```dart
class _EmptyContext extends UnmodifiableMapView<String, Object>
    implements Context {
  const _EmptyContext(): super(const {});
}
```
or without the empty class, but keeping the `Context.empty()` as a constructor and not singleton (let's imagine you need to handle generics even though this example doesn't)

```dart
class Context extends UnmodifiableMapView<String, Object> {
  factory Context(Map<String, Object> from) {
    if (from is Context) return from;
    if (from.isEmpty) return empty;
    return Context._(from);
  }
  static final empty = Context._(const {});
  Context._(Map<String, Object> from) : super(from);
}
```

I don't think I see a reason we'd need this class in this library. I'd be happy to explore whether we can make the existing constructor `const`.

natebosch:
> I'm assuming after null safety lands the package is being bumped to 2.0 right?

No, this and a few other very core packages are migrating to null safety without a major version bump because an early choice by the Flutter framework makes it infeasible to do breaking releases.

mit-mit:
Closing as assumed stale

PR #178, opened 2021-02-01
Prepare to publish stable null safety

natebosch:


natebosch:
https://github.com/dart-lang/collection/compare/dart-lang:eda5243...dart-lang:a79d521

No differences in `lib/`

I expect CI to fail because `test` has an upper bound that disallows `1.15.0` version of `collection`, but is known to be compatible.

jakemac53 (approved):

PR #180, opened 2021-02-02
add mapNotNull

hoc081098:
- #180 
Returns an iterable containing only the non-`null` results
of applying the given [transform] function to each element in this iterable.

### DIFF
--- a/lib/src/iterable_extensions.dart
+++ b/lib/src/iterable_extensions.dart
@@ -169,7 +169,7 @@ extension IterableExtension<T> on Iterable<T> {
   /// Returns an iterable containing only the non-`null` results
   /// of applying the given [transform] function to each element in this iterable.
   Iterable<R> mapNotNull<R extends Object>(R? Function(T) transform) sync* {
-    for (final element in this) {
+    for (var element in this) {
       final value = transform(element);
       if (value != null) {
         yield value;
--- a/lib/src/iterable_extensions.dart
+++ b/lib/src/iterable_extensions.dart
@@ -166,8 +166,13 @@ extension IterableExtension<T> on Iterable<T> {
     }
   }
 
-  /// Returns an iterable containing only the non-`null` results
-  /// of applying the given [transform] function to each element in this iterable.
+  /// The non-`null` results of calling [transform] on the elements of [this].
+  ///
+  /// Returns a lazy iterable which calls [transform]
+  /// on the elements of this iterable in iteration order,
+  /// then emits only the non-`null` values.
+  ///
+  /// If [transform] throws, the iteration is terminated.
   Iterable<R> mapNotNull<R extends Object>(R? Function(T) transform) sync* {
     for (var element in this) {
       final value = transform(element);
--- a/lib/src/iterable_extensions.dart
+++ b/lib/src/iterable_extensions.dart
@@ -175,7 +175,7 @@ extension IterableExtension<T> on Iterable<T> {
   /// If [transform] throws, the iteration is terminated.
   Iterable<R> mapNotNull<R extends Object>(R? Function(T) transform) sync* {
     for (var element in this) {
-      final value = transform(element);
+      var value = transform(element);
       if (value != null) {
         yield value;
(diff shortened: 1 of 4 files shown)


lib/src/iterable_extensions.dart
```diff
@@ -166,6 +166,17 @@ extension IterableExtension<T> on Iterable<T> {
     }
   }
 
+  /// Returns an iterable containing only the non-`null` results
+  /// of applying the given [transform] function to each element in this iterable.
+  Iterable<R> mapNotNull<R extends Object>(R? Function(T) transform) sync* {
```
lrhn, line 171: This is basically a combination of `map` and `whereNotNull`. Or a variant of `expand` where there is always zero or one element in the resulting list.

I've often wanted a variant of `map` with a built-in `where` to skip some elements, and this function would satisfy that.

So, definitely useful.

lib/src/iterable_extensions.dart
```diff
@@ -166,6 +166,17 @@ extension IterableExtension<T> on Iterable<T> {
...
+    for (final element in this) {
```
lrhn, line 172: Just use `var` instead of `final`. This package generally doesn't use `final` for local variables.

lib/src/iterable_extensions.dart
```diff
@@ -166,6 +166,17 @@ extension IterableExtension<T> on Iterable<T> {
     }
   }
 
+  /// Returns an iterable containing only the non-`null` results
```
lrhn, line 169: Documentation should have a shorter first line.
I'd say:

```dart
/// The non-`null` results of calling [transform] on the elements of [this].
///
/// Returns a lazy iterable which calls [transform] 
/// on the elements of this iterable in iteration order,
/// then emits only the non-`null` values.
```
(Generally, I'd document iterables or collections by documenting their elements.)

test/extensions_test.dart
```diff
@@ -243,6 +243,31 @@ void main() {
...
+        test('single', () {
+          expect(iterable(<String>['1']).mapNotNull(int.tryParse), [1]);
+          expect(iterable(<String>['#']).mapNotNull(int.tryParse), isEmpty);
+        });
+        test('multiple', () {
+          expect(iterable(<String>['1', '2', '3']).mapNotNull(int.tryParse),
+              [1, 2, 3]);
+          expect(iterable(<String>['1', '#', '3']).mapNotNull(int.tryParse),
+              [1, 3]);
+          final transform = (int? i) =>
```
lrhn, line 259: Just declare this as a function: `List<String>? transform(int? i) => ...`.

lrhn (approved):
Needs an entry in CHANGELOG.md, but otherwise very good!

hoc081098:
Thanks @lrhn for reviewing, I have already fixed and added an entry in changelog

natebosch:
The name makes me think the behavior would match `.whereNotNull().map(...)`, but it looks like this is `.map(...).whereNotNull`.

I'm not sure I have a better proposal. Maybe `mapAndFilter`?

natebosch:
We should also consider whether it's better to not have this API since the alternative is equally (or more) readable and the perf isn't likely to be meaningfully different for most usages.

lrhn:
Another name could be `optionalMap` or `mapOptional`, because you either map an input value to an output value, or to no value.

That's the functionality I like about it. Together with `map` and `expand`, it completes the combinations:
* .map: S &mapsto; T
* .expand: S &mapsto: T*
* .mapNotNull: S &mapsTo: T?

I do think the functionality deserves to exist by itself.

Naming is hard. Maybe `mapSelect` or `mapOrNot` (nope, not the last one!). 
Actually `mapNotNull` isn't bad.

natebosch:
After some more discussion about the naming and usability of the API I do feel more strongly that we should skip this implementation altogether. I agree that it would be nice to have each of the mapping combinations represented, but the naming difficulty is a good indication that `.map(...).whereNotNull()` is going to be the right choice most of the time.

PR #181, opened 2021-02-22
Removed unnecessary nullable type in maxBy

Afsar-Pasha:


lib/src/functions.dart
```diff
@@ -86,7 +86,7 @@ S? minBy<S, T>(Iterable<S> values, T Function(S) orderBy,
 /// compared using their [Comparable.compareTo].
 ///
 /// Returns `null` if [values] is empty.
-S? maxBy<S, T>(Iterable<S> values, T Function(S?) orderBy,
+S? maxBy<S, T>(Iterable<S> values, T Function(S) orderBy,
```
lrhn, line 89: Good catch.

lrhn (approved):

natebosch:
This change causes errors in google3 and is causing pain trying to get the package rolled through.

@mit-mit - A data point that a cbuild like system for package PRs would be a big help so we find out before it becomes a blocker.

PR #182, opened 2021-03-05
Unorderd list equality

hossameldinmi:
That is used to compare between 2 `List<List<T>>` without considering order.
I found it necessary if I want to use `equals` list extension method.


hossameldinmi:
@googlebot I signed it!

hossameldinmi:
@googlebot I signed it!

lrhn:
This class should not be necessary, as I understand it, since you can already use `UnorderedIterableEquality` for that. Lists are iterables.
Can you say how this class is better than `UnorderedIterableEquality` itself?

PR #184, opened 2021-03-14
Move all extensions methods into their own extensions

### LINKED ISSUE #183: Allow hiding of conflicting extensions
Currently, it is not possible to hide individual extension function which conflict with extensions of other packages. It's only possible to hide a whole "extension groups" such as `IterableExtension`, but not `IterableExtension.sortBy` individually.

Please wrap every extension function in an individually named extension so it can be included/excluded with `hide`/`show`.

Example:

```dart
/// Currently
extension IterableExtension<T> on Iterable<T> {
 
  List<T> sample(int count, [Random? random]) {
    // ...
  }

  Iterable<T> whereNot(bool Function(T element) test) {
    // ...
  }

  List<T> sortedBy<K extends Comparable<K>>(K Function(T element) keyOf) {
    // ...
  }
}
```

```dart
/// Proposal
extension SampleIterableExtension<T> on Iterable<T> {
  List<T> sample(int count, [Random? random]) {
    // ...
  }
}

extension WhereNotIterableExtension<T> on Iterable<T> {
  Iterable<T> whereNot(bool Function(T element) test) {
    // ...
  }
}

extension SortedByIterableExtension<T> on Iterable<T> {
  List<T> sortedBy<K extends Comparable<K>>(K Function(T element) keyOf) {
    // ...
  }
}
```

passsy:
This allows the usage of show/hide to resolve naming conflicts with other libraries

Allows the usage of `package:dartx` and `package:collection` in a single dart file with `show`/`hide`

Fixes #183

### DIFF
--- a/lib/src/iterable_extensions.dart
+++ b/lib/src/iterable_extensions.dart
@@ -8,16 +8,13 @@ import 'package:collection/src/utils.dart';
 
 import 'algorithms.dart';
 
-/// Extensions that apply to all iterables.
-///
-/// These extensions provide direct access to some of the
-/// algorithms expose by this package,
-/// as well as some generally useful convenience methods.
-///
-/// More specialized extension methods that only apply to
-/// iterables with specific element types include those of
-/// [IterableComparableExtension] and [IterableNullableExtension].
+@Deprecated('Use specialized extensions such as IterableSortedExtension')
 extension IterableExtension<T> on Iterable<T> {
+  // Do not add extension methods here.
+  // Place each extension method in its own extension.
+}
+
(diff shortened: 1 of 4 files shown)


lrhn:
If this is the pattern we expect people to use in general (and I don't see why not), I think we should reconsider our extension syntax. If every extension has just one member, the syntax is far too heavy.

This is technically a breaking change. If someone was hiding the old names in order to avoid a conflict, they now get that conflict anyway (until they change the "hide" name). They may also get warnings about hiding an non-existing name.
Are we sure we want to do it in a minor-version increment?

mit-mit:
Closing as assumed stale

PR #192, opened 2021-05-02
Use a stable sort algorithm for Iterable.sortedBy

xvrh:
The random nature of the quick sort algorithm is surprising for an extension like `sortedBy`.

For reference, C# uses a stable sort for [`Enumerable.OrderBy`](https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.orderby?redirectedfrom=MSDN&view=net-5.0#System_Linq_Enumerable_OrderBy__2_System_Collections_Generic_IEnumerable___0__System_Func___0___1__)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,6 @@
-## 1.15.1-dev
+## 1.16.0
+
+* Use a stable sort algorithm in the `IterableExtension.sortedBy` method.
 
 ## 1.15.0
 
--- a/lib/src/iterable_extensions.dart
+++ b/lib/src/iterable_extensions.dart
@@ -64,7 +64,7 @@ extension IterableExtension<T> on Iterable<T> {
   /// property [keyOf] of the element.
   List<T> sortedBy<K extends Comparable<K>>(K Function(T element) keyOf) {
     var elements = [...this];
-    quickSortBy<T, K>(elements, keyOf, compareComparable);
+    mergeSortBy<T, K>(elements, keyOf, compareComparable);
     return elements;
   }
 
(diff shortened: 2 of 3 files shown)


lrhn:
I don't know why people keep expecting stable sorts, but apparently the do.
Looks fine to me.

Do update the version number in pubspec.yaml to the next minor version number and add a CHANGELOG.md entry explaining the change.

tatumizer:
> I don't know why people keep expecting stable sorts, but apparently they do.

Example: suppose you are an innovator composer and you write a suite for 2 sirens. Each siren has a distinct tone controlled by 2 commands: "siren on" and "siren off". Using a computer program written in dart, you generate 2 tracks independently, each track being sorted in ascending order of timestamps. The intervals between "siren on" and "siren off" can sometimes be zero due to rounding, but "siren on" always comes first. When you mix 2 tracks together, you want to preserve the original order of timestamps, otherwise you can end up in a situation where "siren off" comes before "siren on", which will result in a very long note that may disturb the neighbors.

(Example taken from my own practice)

xvrh:
Thanks @lrhn 

> Do update the version number in pubspec.yaml to the next minor version number and add a CHANGELOG.md entry explaining the change.

Done.

lrhn:
LGTM. Thanks!

@tatumizer That's a perfectly good example of why people *who assume* stable sorting will get bitten. It's why you'd assume that to begin with that surprises me (but I guess I've just implemented enough quicksorts to make be immune to that assumption).

Quicksort is faster and more space efficient, so I'm a little sad to see something slower being the default.

lrhn (approved):

tatumizer:
I think every beginner programmer gets bitten by this at least once. It's not that easy to realize what you assume - every misunderstanding comes from some wrong implicit assumption. But WRT sorting, the (implicit) stability assumption comes from a common-sense argument: when we sort something by hand, we never swap equal things; the computer must be smarter than me, hence it won't swap them either. 
(Dont' take my comment too seriously: I'm long past the stage of composing for sirens - moved on to steam whistles already :-)).

PR #200, opened 2021-06-19
Add seperatedBy and toMap extensions.

North101:
Adds `seperatedBy` and `toMap` extensions.

These were extensions I was using in my own projects and thought I'd contribute.

`seperatedBy` example
```dart
return Column(children: list.map((e) => ListTile(title: Text(e)).seperatedBy(const Divider()).toList());
```

`toMap` example
```dart
return list.mapIndexed((index, e) => MapEntry(index, e)).toMap();
```

### DIFF
--- a/lib/src/iterable_extensions.dart
+++ b/lib/src/iterable_extensions.dart
@@ -589,6 +589,19 @@ extension IterableExtension<T> on Iterable<T> {
     }
     return true;
   }
+
+  /// The elements with [seperator] between each.
+  Iterable<T> seperatedBy(T seperator) sync* {
+    if (isEmpty) return;
+
+    final iterator = this.iterator;
+    iterator.moveNext();
+    yield iterator.current;
+    while (iterator.moveNext()) {
+      yield seperator;
+      yield iterator.current;
+    }
+  }
 }
 
 /// Extensions that apply to iterables with a nullable element type.
@@ -832,4 +845,9 @@ extension ComparatorExtension<T> on Comparator<T> {
         if (result == 0) result = tieBreaker(a, b);
         return result;
       };
+}
+
+extension MapEntryExtension<K, V> on Iterable<MapEntry<K, V>> {
+  /// Creates a map from [Iterable<MapEntry>].
+  Map<K, V> toMap() => Map.fromEntries(this);
 }


North101:
@googlebot I signed it!

lib/src/iterable_extensions.dart
```diff
@@ -589,6 +589,19 @@ extension IterableExtension<T> on Iterable<T> {
     }
     return true;
   }
+
+  /// The elements with [seperator] between each.
+  Iterable<T> seperatedBy(T seperator) sync* {
+    if (isEmpty) return;
+
+    final iterator = this.iterator;
+    iterator.moveNext();
```
lrhn, line 598: Should check whether this returns false, and return if it does, instead of having a separate `isEmpty` at the top. (In general, it's best to not nterate the same iterable more than once).

lib/src/iterable_extensions.dart
```diff
@@ -589,6 +589,19 @@ extension IterableExtension<T> on Iterable<T> {
     }
     return true;
   }
+
+  /// The elements with [seperator] between each.
```
lrhn, line 593: This documentation should be more elaborate.
Something like:
```dart
/// The elements of [this] interleaved by [separator].
///
/// Inserts [separator] between any two elements of this iterable.
/// If this iterable is empty or has only one element, no separators are inserted,
/// otherwise a separator is inserted before each element of this iterable
/// other than the first.
```
(so we cover the the edge cases explicitly).

lib/src/iterable_extensions.dart
```diff
@@ -833,3 +846,8 @@ extension ComparatorExtension<T> on Comparator<T> {
         return result;
       };
 }
+
+extension MapEntryExtension<K, V> on Iterable<MapEntry<K, V>> {
```
lrhn, line 850: Should have a comment, even if just
```dart
/// Extension methods on collections of [MapEntry]s.
```

lib/src/iterable_extensions.dart
```diff
@@ -833,3 +846,8 @@ extension ComparatorExtension<T> on Comparator<T> {
...
+  /// Creates a map from [Iterable<MapEntry>].
```
lrhn, line 851:
```dart
/// Creates a map containing the entries described by [this].
///
/// Returns a `Map<K, V>`, as if created by `<K, V>{}`
/// and then having each key/value pair from [this] added to the map in 
/// iteration order (so a later entry with the same key as an earlier entry will 
/// overwrite the earlier value).

**Review by lrhn — COMMENTED** [review]
The functionality is reasonable and probably useful (not absolutely sure about the `separatedBy`, but that's because I can't imagine a use-case for it other than comma-separating a list of strings.) 

It does need tests and a CHANGELOG entry.

File: lib/src/iterable_extensions.dart [hunk 5]
```diff
@@ -589,6 +589,19 @@ extension IterableExtension<T> on Iterable<T> {
     }
     return true;
   }
+
+  /// The elements with [seperator] between each.
+  Iterable<T> seperatedBy(T seperator) sync* {
```
natebosch, line 594: This one feels pretty specialized. I'm not sure if there will be wide enough usage to have this in a core package.

lrhn: It worries me a little too. I have seen other versions of it in wild, one called `interspersed` and used to insert padding between Flutter widgets in a list.
It's basically the `join` function separator, without the actual joining.

I'm guessing one use-case could be something where you do a `for`-`in` on the result, and treat the separators specially, so you can "functionally" handle the fence-post problem.

So yes, I do want to hear use-cases for `separatedBy` before agreeing to it.

natebosch:
Closing for now.

If you still have a need for `separatedBy` we can discuss the use case and whether it should get shipped in this package.

Issue #208, opened 2021-09-03
DeepCollectionEquality is not symmetric most of the time 


jakemac53:
When comparing a List/Set to an Iterable, the order of the arguments will give a different result, making them non-symmetric.

This actually varies on the implementation though - in the case of `DeepCollectionEquality.unordered` then the results are symmetric, but only when comparing specifically a List and an Iterable 😱 . 

Fixing this would be breaking at this point. Some options available to us are:

- Change the behavior and do a breaking change.
- Change the behavior but consider it as a bug fix and don't do a breaking change.
- Deprecate existing constructors, add new ones that behave in a more sane (or at least consistent) manner.
- Add new constructors that behave consistently, leave old ones in place and don't deprecate.
- Do nothing.

cc @natebosch @lrhn

jakemac53:
See the tests in the linked pr (https://github.com/dart-lang/collection/pull/209) which demonstrate the behavior.

natebosch:
I think I would be in favor of writing a symmetric `CollectionEquality` and deprecating `DeepCollectionEquality`.

jakemac53:
I am not sure that `CollectionEquality` implies deep equality though? Unless we wanted to make whether it uses deep equality a configurable thing?

PR #209, opened 2021-09-03
update deep collection tests to show the state of the world re:symmetric equality

jakemac53:
This just demonstrates the current state of the world here. We should separately investigate if we want to change the behavior or not.


jakemac53:
Related to https://github.com/dart-lang/collection/issues/208

natebosch (approved):

PR #210, opened 2021-09-06
Add changelog entry for BoolList.

lrhn:
Add author to AUTHORS.
Small cleanups.
Rename `.from` constructor to `.of`, for consistency.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -4,3 +4,4 @@
 #   Name/Organization <email address>
 
 Google Inc.
+AAABramenko (https://github.com/AAAbramenko)
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -1,5 +1,5 @@
 name: collection
-version: 1.16.0
+version: 1.17.0-dev
 
 description: Collections and utilities functions and classes related to collections.
 repository: https://github.com/dart-lang/collection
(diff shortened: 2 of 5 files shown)


AUTHORS
```diff
@@ -4,3 +4,4 @@
 #   Name/Organization <email address>
 
 Google Inc.
+AAABramenko (https://github.com/AAAbramenko)
```
lrhn, line 7: @AAAbramenko Do you want an entry in this file, or would you prefer not to? Do you want your e-mail listed here or just your github link?

AAAbramenko: I'll be glad. I think nickname and github link are enough.

pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: collection
-version: 1.16.0
+version: 1.17.0-dev
```
natebosch, line 2: We are still on `1.16.0`.

https://github.com/dart-lang/collection/pull/211

natebosch (approved):

pubspec.yaml
lrhn: ACK.

natebosch (approved):

PR #212, opened 2021-09-08
fix #189 (.max, .maxOrNull, .min, .minOrNull) for Iterable<int> and Iterable<double>

### LINKED ISSUE #189: .min/max extension not working on List<int>
It's surprising that `.min` and `.max` extensions work on a static type `List<num>` but not with `List<int/double>`

```dart
import 'package:collection/collection.dart';

void main() {
  print(<num>[1, 2].min); // OK
  print([1, 2].min); // Static error
  print(<double>[1.0, 2.0].min); // Static error
}
```

TimWhiting:
`Iterable<int>` and `Iterable<double>` are not covered by the current `IterableComparableExtension` because they both implement `Comparable<num>` See Issue #189. This fixes that issue and adds tests.


kevmoo:
Should have a changelog here quick!

TimWhiting:
@kevmoo 
Do you want me to add the changelog here or let @lrhn add it to #210?

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## 1.16.0-dev
 
 * Use a stable sort algorithm in the `IterableExtension.sortedBy` method.
+* Add `min` `max` `minOrNull` and `maxOrNull` getters to `IterableDoubleExtension` and `IterableIntegerExtension`
```
lrhn, line 4: Commas after \`min\` and \`max\`.

lib/src/iterable_extensions.dart
```diff
@@ -601,6 +601,74 @@ extension IterableNumberExtension on Iterable<num> {
...
+      }
+      return value;
+    }
+    return null;
+  }
+
+  /// A minimal element of the iterable.
+  ///
+  /// The iterable must not be empty.
+  int get min {
```
lrhn, line 623: Can be shortened to 
```dart
 int get min => minOrNull ?? throw StateError("No element");
```
Ditto for `max` below.

lib/src/iterable_extensions.dart
```diff
@@ -643,6 +711,74 @@ extension IterableIntegerExtension on Iterable<int> {
 ///
 /// Specialized version of some extensions of [IterableNumberExtension].
 extension IterableDoubleExtension on Iterable<double> {
+  /// A minimal element of the iterable, or `null` it the iterable is empty.
```
lrhn, line 714: Should we handle NaNspecially? I think we should. If any element is NaN, we should return NaN, both for `min` and `max`.

We can insert
```dart
   if (value.isNaN) return value;
    ... 
     if (newValue.isNaN) return newValue;
```
after reading values from the iterator.

(I wouldn't use the behavior of `double.compareTo`, that's intended to be a complete ordering for *sorting*, and therefore puts it at one of the ends - can't remember which - but here I think returning NaN is the always right thing to do if any value is NaN).

lib/src/iterable_extensions.dart
```diff
@@ -643,6 +711,74 @@ extension IterableIntegerExtension on Iterable<int> {
...
+      }
+      return value;
+    }
+    return null;
+  }
+
+  /// A minimal element of the iterable.
+  ///
+  /// The iterable must not be empty.
+  double get min {
```
lrhn, line 733: Again `=> minOrNull ?? throw StateError("No element");` and ditto for max.

AUTHORS
```diff
@@ -4,3 +4,4 @@
 #   Name/Organization <email address>
 
 Google Inc.
+TimWhiting tim@whitings.org
```
lrhn, line 7: Add a trailing newline.

test/extensions_test.dart
```diff
@@ -887,6 +887,74 @@ void main() {
           expect(iterable(<num>[1, 3, 5, 9]).average, 4.5);
         });
       });
+      group('.min', () {
+        test('empty', () {
+          expect(() => iterable(<int>[]).min, throwsStateError);
+          expect(() => iterable(<double>[]).min, throwsStateError);
+          expect(() => iterable(<num>[]).min, throwsStateError);
```
lrhn, line 894: Where does `Iterable<num>.min` come from?

lib/src/iterable_extensions.dart
```diff
@@ -601,6 +601,74 @@ extension IterableNumberExtension on Iterable<num> {
 ///
 /// Specialized version of some extensions of [IterableNumberExtension].
```
lrhn, line 602: Maybe even add:
```
///  ... ,
/// or [IterableComparableExtension] since integers are only `Comparable<num>`.

**Review by lrhn — CHANGES_REQUESTED** [review]
Nits only, except for handling NaN for doubles, code looks good!

File: CHANGELOG.md (hunk 1 above)
Comment (TimWhiting) [reply CHANGELOG.md:4]: Done

File: lib/src/iterable_extensions.dart (hunk 3 above)
Comment (TimWhiting) [reply lib/src/iterable_extensions.dart:714]: Done

File: lib/src/iterable_extensions.dart (hunk 4 above)
Comment (TimWhiting) [reply lib/src/iterable_extensions.dart:733]: See above

File: AUTHORS (hunk 5 above)
Comment (TimWhiting) [reply AUTHORS:7]: Done

File: test/extensions_test.dart (hunk 6 above)
Comment (TimWhiting) [reply test/extensions_test.dart:894]: IterableComparableExtension, updated that to also simplify the non-null cases to proxy to the nullable getter.

File: lib/src/iterable_extensions.dart (hunk 7 above)
Comment (TimWhiting) [reply lib/src/iterable_extensions.dart:602]: Done.

File: lib/src/iterable_extensions.dart (hunk 2 above)
Comment (TimWhiting) [reply lib/src/iterable_extensions.dart:623]: Tried this but it complains that it expects an identifier where throw is. But I updated it to delegate to minOrNull / maxOrNull now.
expects_identifier

File: lib/src/iterable_extensions.dart (hunk 3 above)
Comment (TimWhiting) [reply lib/src/iterable_extensions.dart:702]: Added tests for this case as well.

Comment (TimWhiting) [reply lib/src/iterable_extensions.dart:702]: So, I was about to add tests for `Iterable<num>` but since it uses `compareTo` it doesn't return NaN for `.minOrNull`. If I add an extension on Iterable<num> it will conflict with the IterableComparable definition.

**Comment by TimWhiting** [comment]
@lrhn 
Ready for re-review. I had one question about the NaN.

File: lib/src/iterable_extensions.dart (hunk 2 above)
Comment (lrhn) [reply lib/src/iterable_extensions.dart:624]: My bad, I keep forgetting the precedence of `throw`. It can be written as:
```dart
  => maxOrNull ?? (throw StateErrro('No element');
```
(The error message is confusing because other things than identifiers can also occur, it basically means that it expects an expression with a different precedence).

lib/src/iterable_extensions.dart
lrhn: An iterable on `num` should take precedence over the one on `Iterable<T = num>`.

Since `num` is a proper subtype of `Iterable<num>`, it's a more specific type and the extension conflict resolution should prefer `on Iterable<num>` to `on Iterable<Comparable<num>>`.

It's technically a breaking change to change the behavior of `Iterable<num>`, but we are already doing that for `Iterable<double>` (which would previously also match `Iterable<Comparable<num>>` and return a `num`).

I'd add the `num` version too, and handle `NaN` there as well.

lib/src/iterable_extensions.dart
lrhn: Yep, needs parentheses.

test/extensions_test.dart
lrhn: Sweet!

lib/src/iterable_extensions.dart
```diff
@@ -730,18 +844,11 @@ extension IterableComparableExtension<T extends Comparable<T>> on Iterable<T> {
...
-        if (value.compareTo(newValue) < 0) {
-          value = newValue;
-        }
-      }
-      return value;
+    final result = maxOrNull;
+    if (result == null) {
+      throw StateError('No element');
     }
-    throw StateError('No element');
```
lrhn, line 744: Yes, this should be safe because a nullable type cannot possibly implement `Comparable<T>` for any `T`.

lrhn (approved):

TimWhiting:
@lrhn Ready for merge

lrhn (approved):
Very nice, thank you!

PR #213, opened 2021-09-12
Fix the splitBetween example in the doc

fieldinrain:
In the splitBetween, the test function has not the index parameter `i`.


fieldinrain:
@googlebot I signed it!

lrhn (approved):
Good catch, thanks!

PR #215, opened 2021-09-14
Drop nullability from _UnorderedEquality generic

### LINKED ISSUE #214: Equality subclasses unnecessarily allow null arguments
When we migrated to null safety we introduced a discrepancy and unnecessarily expanded the argument types for `equals` and `hash`.

We declare most of the classes to `implements Equality<T>`, but allow a `T?` for `equals` and `hash`, so when they are used directly they allow `null` arguments, but when used as an `Equality` the functionality is unnecessary.

For `_UnorderedEquality` this came from making `T` unnecessarily nullable,  and in other subclasses the class level generics are fine but the method arguments are still expanded to allow `null`.

This entire test should be static errors in my opinion:
https://github.com/dart-lang/collection/blob/cd46e1959d1051aeba9c372de4f59cc20cb358e6/test/equality_test.dart#L243-L267

Changing the arguments to non-nullable (with some extra handling of the nulls that flow from libraries that have not migrated to null safety) is a better design, but the impact of the breaking change is likely higher.

Changing the implements to `implements Equality<T?>` gives a more accurate interface and expands the places these classes can be used, but is still breaking.

See https://github.com/dart-lang/collection/pull/161#discussion_r510213398

natebosch:
Closes #214

The nullable `T` in `Equality<T>` changes the assignability and
introduces a difference between the other collection equality
implementations and the unordered implementations.

Prior to this change `SetEquality<T>` could not be statically assigned
to a `Equality<Set<T>>`, instead it had to be assigned to
`Equality<Set<T>?>`, while `ListEquality<T>` was assignable to
`Equality<List<T>>`. Now both are assignable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 * Add `BoolList` class for space-efficient lists of boolean values.
 * Use a stable sort algorithm in the `IterableExtension.sortedBy` method.
 * Add `min`, `max`, `minOrNull` and `maxOrNull` getters to `IterableDoubleExtension`, `IterableNumberExtension` and `IterableIntegerExtension`
+* Change `UnorderedIterableEquality` and `SetEquality` to implement `Equality`
+  with a non-nullable generic. This allows assignment to `Equality` with a
+  non-nullable generic.
+
 
 ## 1.15.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,14 @@
 
 * Add `BoolList` class for space-efficient lists of boolean values.
 * Use a stable sort algorithm in the `IterableExtension.sortedBy` method.
-* Add `min`, `max`, `minOrNull` and `maxOrNull` getters to `IterableDoubleExtension`, `IterableNumberExtension` and `IterableIntegerExtension`
+* Add `min`, `max`, `minOrNull` and `maxOrNull` getters to
+  `IterableDoubleExtension`, `IterableNumberExtension` and
+  `IterableIntegerExtension`
 * Change `UnorderedIterableEquality` and `SetEquality` to implement `Equality`
-  with a non-nullable generic. This allows assignment to `Equality` with a
-  non-nullable generic.
+  with a non-nullable generic to allows assignment to variables with that type.
+  Assignment to `Equality` with a nullable type is still allowed because of
+  covariance. The `equals` and `hash` methods continue to accept nullable
+  arguments.
 
 
 ## 1.15.0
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,10 @@
 * Add `BoolList` class for space-efficient lists of boolean values.
 * Use a stable sort algorithm in the `IterableExtension.sortedBy` method.
 * Add `min`, `max`, `minOrNull` and `maxOrNull` getters to `IterableDoubleExtension`, `IterableNumberExtension` and `IterableIntegerExtension`
+* Change `UnorderedIterableEquality` and `SetEquality` to implement `Equality`
+  with a non-nullable generic. This allows assignment to `Equality` with a
+  non-nullable generic.
```
lrhn, line 8: This is a breaking change, you can no longer use it as equality for `Set<E>?`s.

This is the change I was talking about doing yesterday, but I think it needs to be a major version increment.
(Also, I'd add `class NullableEquality<E extends Object> implements Equality<E?>` which extends equality to from a type to the nullable type by equalting `null` only with itself).

lib/src/equality.dart
```diff
@@ -202,14 +202,14 @@ class ListEquality<E> implements Equality<List<E>> {
...
-abstract class _UnorderedEquality<E, T extends Iterable<E>?>
+abstract class _UnorderedEquality<E, T extends Iterable<E>>
     implements Equality<T> {
   final Equality<E> _elementEquality;
 
   const _UnorderedEquality(this._elementEquality);
 
   @override
-  bool equals(T elements1, T elements2) {
+  bool equals(T? elements1, T? elements2) {
```
lrhn, line 212: Adding the `?` here does fix the backwards compatibility issue.
I guess it's sufficient as long as you don't cast to `Equality<Set<E>>`, and if you did, it wouldn't work anyway.

I still want to remove this `?` too, but that'll likely require a major version bump.

lrhn (approved):

lib/src/equality.dart
natebosch: > I still want to remove this `?` too, but that'll likely require a major version bump.

Yeah I think that would have been a better migration. We can keep #214 open and tag it for the next major version release.

> We're also inconsistent in that other equalities have not been extended to nullable types).

`IterableEquality`, `MapEquality`, and `ListEquality` all have this same pattern, they accept nullable arguments but the declared `Equality` interface is with a non-nullable type.

lrhn: Ah yes. We are consistently weird :)

LGTM!

PR #217, opened 2021-09-17
Add Iterable.elementAtOrNull extension

windrunner414:
Add a extension function for List `getOrNull(int index)`
returns a element at the given [index] of this list, if [index] out of bounds, return `null`

Example:
```dart
final list = [1,2,3];
list.getOrNull(1); // 2
list.getOrNull(10); // null
```

### DIFF
--- a/lib/src/iterable_extensions.dart
+++ b/lib/src/iterable_extensions.dart
@@ -352,10 +352,16 @@ extension IterableExtension<T> on Iterable<T> {
     }
     return null;
   }
-  
-  /// Returns an element at the given [index]
-  /// or `null` if the [index] is out of bounds of this list.
-  /// [index] must not be negative
+
+  /// The [index]th element, or `null` if there is no such element.
+  ///
+  /// Returns the element at position [index] of this iterable,
+  /// just like [elementAt], if this iterable has such an element.
+  /// If this iterable does not have enough elements to have one with the given
+  /// [index], the `null` value is returned, unlike [elementAt] which throws
+  /// instead.
+  ///
+  /// The [index] must not be negative.
   T? elementAtOrNull(int index) => skip(index).firstOrNull;
 
   /// Groups elements by [keyOf] then folds the elements in each group.
--- a/lib/src/iterable_extensions.dart
+++ b/lib/src/iterable_extensions.dart
@@ -352,6 +352,11 @@ extension IterableExtension<T> on Iterable<T> {
     }
     return null;
   }
+  
+  /// Returns an element at the given [index]
+  /// or `null` if the [index] is out of bounds of this list.
+  /// [index] must not be negative
+  T? elementAtOrNull(int index) => skip(index).firstOrNull;
 
   /// Groups elements by [keyOf] then folds the elements in each group.
   ///
--- a/lib/src/list_extensions.dart
+++ b/lib/src/list_extensions.dart
@@ -258,10 +258,17 @@ extension ListExtensions<E> on List<E> {
     }
     return true;
   }
-  /// Returns an element at the given [index]
-  /// or `null` if the [index] is out of bounds of this list.
-  /// [index] must not be negative
-  T? elementAtOrNull(int index) => (index < length) ? this[index] : null;
+
+  /// The [index]th element, or `null` if there is no such element.
+  ///
+  /// Returns the element at position [index] of this list,
+  /// just like [elementAt], if this list has such an element.
+  /// If this list does not have enough elements to have one with the given
+  /// [index], the `null` value is returned, unlike [elementAt] which throws
+  /// instead.
+  ///
+  /// The [index] must not be negative.
+  E? elementAtOrNull(int index) => (index < length) ? this[index] : null;
 
   /// Contiguous [slice]s of [this] with the given [length].
   ///
--- a/lib/src/list_extensions.dart
+++ b/lib/src/list_extensions.dart
@@ -255,7 +255,8 @@ extension ListExtensions<E> on List<E> {
   
   /// Returns an element at the given [index]
   /// or `null` if the [index] is out of bounds of this list.
-  E? getOrNull(int index) => (index >= 0 && index < length) ? this[index] : null;
+  /// [index] must not be negative
+  T? elementAtOrNull(int index) => (index < length) ? this[index] : null;
 }
 
 /// Various extensions on lists of comparable elements.
--- a/lib/src/list_extensions.dart
+++ b/lib/src/list_extensions.dart
@@ -252,6 +252,10 @@ extension ListExtensions<E> on List<E> {
     }
     return true;
   }
+  
+  /// Returns an element at the given [index]
+  /// or `null` if the [index] is out of bounds of this list.
+  E? getOrNull(int index) => (index >= 0 && index < length) ? this[index] : null;
 }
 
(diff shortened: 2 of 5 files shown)


kevmoo:
At a minimum, we should have a test here.

Not sure about the overall usefulness of this extension, though. CC @natebosch

rakudrama:
I don't think the method is sufficiently useful.
Having indexes be out of bounds is often a symptom of some logical error elsewhere.
```
    list.getNoOrNull(i)
```
has the same null-on-out-of-bounds behaviour as
```
    list.asMap()[i]
```
but the latter is shorter.

windrunner414:
I think `asMap()` has more expensive, if the list is very big.
with nnbd, we must check if the return value is null, so it can't hide the logical error (we know the index may be out of bounds so we use it)
one case is I use list to pass the arguments between javascript and dart in webview, and some positional arguments is optional.`args.getOrNull(2) ?? true` is more convenience
and the other case is use it with `split`, like: `'a,b,c'.split(',').getOrNull(3) ?? 'd'`

lib/src/list_extensions.dart
```diff
@@ -252,6 +252,10 @@ extension ListExtensions<E> on List<E> {
     }
     return true;
   }
+  
+  /// Returns an element at the given [index]
+  /// or `null` if the [index] is out of bounds of this list.
+  E? getOrNull(int index) => (index >= 0 && index < length) ? this[index] : null;
```
lrhn, line 258: Not sure I think this functionality is worth it. Too specialized. No clear use-case.

It allows you to do `list.getOrNull(n) ?? (throw ArgumentError("bad index")` instead of checking the index, but only if the list's element type is not null. It also easily hides errors in an index computation.

windrunner414: `??` to hide errors in an index computation is easier than `index < list.length ? list[index] : value` but I think `list.getOrNull(index) ?? value` is also clear that index might out of bounds.
One problem is that it may hide errors if used to determine whether the element is null. In that case, we should use `list[index] == null` instead of `getOrNull` unless we need to deal with non-existent or null

rakudrama:
> I think `asMap()` has more expensive, if the list is very big.

`asMap()` returns a [`ListMapView`](https://github.com/dart-lang/sdk/blob/main/sdk/lib/internal/list.dart#L223). It is a view of the list, not a copy of the data, so the cost is independent of the list size.

yuxiang-projz:
> > I think `asMap()` has more expensive, if the list is very big.
> 
> `asMap()` returns a [`ListMapView`](https://github.com/dart-lang/sdk/blob/main/sdk/lib/internal/list.dart#L223). It is a view of the list, not a copy of the data, so the cost is independent of the list size.

I see.I mistakenly think that only the constant list will use ListMapView.I find that ListMapView is always used.

windrunner414:
> > I think `asMap()` has more expensive, if the list is very big.
> 
> `asMap()` returns a [`ListMapView`](https://github.com/dart-lang/sdk/blob/main/sdk/lib/internal/list.dart#L223). It is a view of the list, not a copy of the data, so the cost is independent of the list size.

I see.I mistakenly think that only the constant list will use ListMapView.I find that ListMapView is always used.

natebosch:
+1 on this being too specialized to put into a core package.

If we do decide to land this functionality it should be called `elementAtOrNull` to match the existing method and to avoid using the ambiguous term "get".

https://api.dart.dev/stable/2.14.2/dart-core/Iterable/elementAt.html
https://dart.dev/guides/language/effective-dart/design#avoid-starting-a-method-name-with-get

windrunner414:
`list.asMap()[0]` maybe a good solution instead of it.Even though it still has some extra costs.
One thing I think is that elementAtOrNull and firstOrNull are similar.Maybe it would be better to extend on iterable.
.length()/.elementAt may not efficient, and asMap() is impossible.

lrhn:
The `firstOrNull` and `lastOrNull` work well because the only failure mode is the list being empty, and therefore there being no first/last element. They also work well on all iterables.

An `elementAtOrNull(int index)` is more complicated. Not a lot, but still a little. You can fail to have the element by passing too big a number, but also if you pass a negative number. I think the latter should still throw. Then `elementAtOrNull(index)` is just equivalent to `.skip(index).firstOrNull`.
(The `.asMap()[index]` only works for lists, `elementAtOrNull` should work for any iterable).

So, I'm still hesitant to whether it's really needed, but if so, an extension on `Iterable<E>` like `E elementAtOrNull(int index) => skip(RangeError.checkNotNegative(index)).firstOrNull;` would make the most sense. Possibly with optimized versions on `List`/`Set`/`Queue` or other classes with efficient length or direct indexing.

windrunner414:
@lrhn I just add `elementAtOrNull` for Iterable and provide a optimized version on `List`.I'm not sure it makes sense to add this method to `Set`. And we don't need `checkNotNegative` because `skip` and `[]` will check it

lib/src/iterable_extensions.dart
```diff
@@ -352,6 +352,11 @@ extension IterableExtension<T> on Iterable<T> {
     }
     return null;
   }
+  
+  /// Returns an element at the given [index]
+  /// or `null` if the [index] is out of bounds of this list.
+  /// [index] must not be negative
+  T? elementAtOrNull(int index) => skip(index).firstOrNull;
```
lrhn, line 359: That's awesome code reuse. Probably reasonable performance, otherwise we can optimize.

lib/src/iterable_extensions.dart
```diff
@@ -352,6 +352,11 @@ extension IterableExtension<T> on Iterable<T> {
     }
     return null;
   }
+  
+  /// Returns an element at the given [index]
```
lrhn, line 356: Going to nitpick here.

The first paragraph of the dartdoc should preferably fit on a single line.

I'd drop the `Returns` (I personally find that it's never the best word to start a DartDoc, and, yes, I know `elementAt` does start with `Returns`).

Maybe
```dart
/// The [index]th element, or `null` if there is no such element.
///
/// Returns the element at position [index] of this iterable, 
/// just like [elementAt], if the iterable has such an element.
/// If this iterable does not have enough elements to have
/// one with the given [index], the `null` value is returned instead,
/// unlike [elementAt] which throws instead.
///
/// The [index] must not be negative.
```

lrhn:
(This fells like something I'd want to solve generally, instead of introducing extension methods for each possible way to access a list.

Maybe we could create a `List<T>` wrapper, implementing `List<T?>`, which doesn't throw on invalid indices or similar state errors, but returns `null` instead. It would still have a `length`, just wouldn't check it as much, and would reject any modifications through the wrapper.
Or, better, wait for View-types and have a zero-overhead wrapper.)

Until then, this works.

natebosch:
> Until then, this works.

Do you mean you'd like to go forward with this change? Should I start getting it ready to land?

lrhn:
Yes, it's good. Get ready to land!

lrhn:
Looks fine, just need an entry in CHANGELOG, maybe an increment in version number.

pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: collection
-version: 1.16.1-dev
```
lrhn, line 2: It's OK to go directly to 1.17.0 and publish.

PR #218, opened 2021-09-17
Remove some unnecessary `return null;`

natebosch:
There is an implicit null return at the end of the method. With null
safety it's less critical to be explicit about the null return since the
return type makes it clear.

This is the same refactoring as was made in one instance of the pattern
in
https://github.com/dart-lang/collection/pull/212/files#diff-013b70bbf26f849f2a314b4b36ec2d37374b4c8a3e21d1fb5b3d1df9788996d2L692

### DIFF
--- a/lib/src/iterable_extensions.dart
+++ b/lib/src/iterable_extensions.dart
@@ -590,7 +590,6 @@ extension IterableNumberExtension on Iterable<num> {
       }
       return value;
     }
-    return null;
   }
 
   /// A minimal element of the iterable.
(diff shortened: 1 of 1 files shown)


lib/src/iterable_extensions.dart
```diff
@@ -590,7 +590,6 @@ extension IterableNumberExtension on Iterable<num> {
       }
       return value;
     }
-    return null;
```
lrhn, line 593: Please do not make this change.

I personally think the missing `return null;` at the end should be a compile-time error (and will work towards making it so).

If it's a compile-time error to have a `return;` with no value, it should not be allowed to have an explicit fallthrough.

PR #219, opened 2021-09-20
Restore an explicit null return

natebosch:
This was unnecessarily removed in an unrelated change.

https://github.com/dart-lang/collection/pull/218#discussion_r712148802


lrhn (approved):
Thanks, I missed that one :(

PR #224, opened 2021-11-18
fix maxBy compare param signature

### LINKED ISSUE #188: incorrect nullable type in maxBy
this is the `compare` parameter in `maxBy`
https://github.com/dart-lang/collection/blob/7a7cc0aa7407ab4cfbb3ad58abf2d408c6d2680a/lib/src/functions.dart#L90
it should be `int Function(T, T)? compare`

ezamagni:
a tiny change to `maxBy` function signature in order to fix #188


lrhn:
Good fix!

lrhn (approved):

Issue #207, opened 2021-09-02
Add Python-style range() function and iterable


Sunbreak:
Ref: https://github.com/dart-lang/sdk/issues/7775

```dart
extension IntRange on int {
  /// A arithmetic sequence of numbers from [this] to, but not including [end].
  /// 
  /// The difference between two consecutive numbers is [stepSize].
  /// The [stepSize] must be positive (not negative or zero).
  ///
  /// The first number is always [this].
  /// If [this] is smaller than [end], the next number in the sequence is found by
  /// adding [stepSize] to the current number until reaching or exceeding [end].
  /// If [end] is smaller than [this], the next number in the sequence is found by
  /// subtracting [stepSize] from the current number until reaching or subceeding [end].
  Iterable<int> to(int end, [int stepSize = 1]) sync* {
    RangeError.checkValueInInterval(stepSize, 1, null, "stepSize");
    if (start <= end) {
      for (var i = this; i < end; i += stepSize) yield i;
    } else {
      for (var i = this; i > end; i -= stepSize) yield i;
    }
  }
}
```

lrhn:
This is definitely the right *place* for such an extension, should we want to add it, since it creates an iterable.

We've considered such "range" functions on `int` before, and it's always been a problem that it's unclear whether the end point is included or not. Adding *both* to `int` seemed overkill.

With extensions we could possibly have both, say `2.to(4)` and `2.until(5)`.

The implementation is not particularly efficient, unless the compiler recognizes and optimizes it.
You'd probably always be signficantly better off, performance wise, doing `for (var i = start; i < end; i += step) ...` than `for (var i in start.until(end)) ...`.

That's what worries me the most: Adding this might make people think using it is *better* than a normal `for` loop. It won't be. Shorter, possibly, but definitely slower.

Which prompts the question: When would you use such an iterable?
The `for` loop is the place I usually see suggested, and that's exactly where I wouldn't recommend using it.
The other example in [#7775](https://github.com/dart-lang/sdk/issues/7775) was `List.from(start.until(end))`, where you should now write the more efficient `[for (var i = start; i < end; i++) i]`.

So, in the currently proposed use-cases, I wouldn't actually recommend using this feature.
(Again, not unless we promise that compiles will convert `for (var i in start.until(end))` into `for (var i = start, n = end; i < n; i++)` - and that only gets more complicated if we allow stepping down as well as up.)


-----
[...]

PR #228, opened 2021-12-30
int extension

thomassth:
Similar to #207

Since https://github.com/felixblaschke/supercharged is stopping active development, I found the `rangeTo` and `until` within are both very useful tool to generate numbered iterates quickly, and brings dart to feature parity with other languages. I think they should be included to this package.

### DIFF
--- a/lib/src/int_extensions.dart
+++ b/lib/src/int_extensions.dart
@@ -1,0 +1,40 @@
+/// Extensions that apply to all integers.
+///
+/// These extensions provide some generally useful convenience methods.
+/// Credits to supercharged_dart
+extension IntRange on int {
+  /// Creates an [Iterable<int>] that contains all values from current integer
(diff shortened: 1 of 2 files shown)


kevmoo:
CC @lrhn – not sure how I feel about this.

At a minimum, should have tests!

thomassth:
I've added tests now.

lrhn:
I'm not totally sold on adding this functionality.

If we do, the naming should be consistent (one is called `rangeTo`, the other `until`, so either both or neither should use `range`), and the implementation might need some tweaking. I'd prefer a an `Iterator` class which can be inlined by an optimizing compiler where possible, and not allocating a function value for each.

So, I'd say no to this particular implementation, and I'll consider whether we want the functionality or not.
(Discussion should go to #207).

PR #229, opened 2022-01-04
Added sort order helpers

jopmiddelkamp:


natebosch:
I think this is too specialized - if we were to move forward with this we'd expand the names to `sortAscending` and `sortDescending`, at which point the manual function literal passed to `sort` compares pretty favorably.

PR #234, opened 2022-03-08
Add a comment about NaN in min/max extensions

slovnicki:
I noticed that NaN will be both min and max, which might be confusing to the user if not documented.

I do think returning NaNs **is better** approach than the other 2 alternatives I can think of:

- Ignoring them, which will not let user know they have NaNs in the iterable.
- Relying on `false` from comparisons to NaN, which will lead to inconsistent returns depending on NaN's placement (e.g. first or not first) in the iterable.

### DIFF
--- a/lib/src/iterable_extensions.dart
+++ b/lib/src/iterable_extensions.dart
@@ -593,7 +593,7 @@ extension IterableNullableExtension<T extends Object> on Iterable<T?> {
 extension IterableNumberExtension on Iterable<num> {
   /// A minimal element of the iterable, or `null` it the iterable is empty.
   ///
-  /// If an element is NaN, it is considered to be a minimal element.
+  /// If any element is [NaN](double.nan), the result is NaN.
   num? get minOrNull {
     var iterator = this.iterator;
     if (iterator.moveNext()) {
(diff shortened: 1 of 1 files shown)


lib/src/iterable_extensions.dart
```diff
@@ -592,6 +592,8 @@ extension IterableNullableExtension<T extends Object> on Iterable<T?> {
 /// since doubles require special handling of [double.nan].
 extension IterableNumberExtension on Iterable<num> {
   /// A minimal element of the iterable, or `null` it the iterable is empty.
+  ///
+  /// If an element is NaN, it is considered to be a minimal element.
```
lrhn, line 596: I'd phrase this as:
```dart
/// If any element is [NaN](double.nan), the result is NaN.
```
It's not considering NaN to be minimal, rather it's about not even being able to answer the question of which value is minimal if one of the values is not a number. That's the general rule of "if any operand is NaN, the result is NaN" that applies to all double operations.

Ditto everywhere below.

lrhn, line 596: Hmm, pretty over-engineered code, now that I look at it again.
```dart
  num? get minOrNull {
    var empty = true;
    var min = double.infinity;
    for (var value in this) {
       empty = false;
       if (value < min) min = value;
    }
    if (empty) return null;
    return min;
  }
```
seems much more optimizable.

(Maybe I should make a benchmark :grin:)

lrhn (approved):
LGTM with the different phrasing.

lib/src/iterable_extensions.dart
slovnicki: Ah, good point. Thank you for the explanation :+1: 
I pushed the changes.

lrhn (approved):
Looks good to me. Thanks!

PR #244, opened 2022-08-30
Add a List<E>.windowed extension

Kernald:
Prior art: [Kotlin](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/windowed.html)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 * Add a top-level `lastBy()` function that converts an `Iterable` to a `Map` by
   grouping its elements using a function, keeping the last element for each
   computed key. Also available as an extension method on `Iterable`.
+* Add an extension `windowed()` on `List`. Similarly to `slices()`, it gives a
+  chunked view of the list.
 
 ## 1.16.0
 
--- a/lib/src/list_extensions.dart
+++ b/lib/src/list_extensions.dart
@@ -286,6 +286,41 @@ extension ListExtensions<E> on List<E> {
       yield slice(i, min(i + length, this.length));
     }
   }
+
+  /// Windowed view of [this] with the given [length], separated by [step]
+  /// elements.
+  ///
+  /// Each window is a view of this list [length] elements long, except for the
+  /// last ones which may be shorter if [partialWindows] is set to `true`.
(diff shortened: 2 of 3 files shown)


lrhn:
Seems doable.

```dart
extension ListWindowsExtension<E> on List<E> {
   Iterable<ListSlice<E>> slices(int size, {int start = 0, int? end, int step = 1}) sync* {
     end = RangeError.checkValidRange(start, end, this);
     for (var i = start; i < end; i += step) {
       var sliceEnd = i + step; 
       if (sliceEnd > end) sliceEnd = end;
       yield this.slice(i, sliceEnd);
     }
   }
}
```

Kernald:
I think `var sliceEnd = i step;` should be `var sliceEnd = i + size`? But otherwise it looks like it should work indeed, from a quick glance.

As a side note, I'm on holidays for a couple of weeks without access to a computer, so I won't be able to address any comments here, but feel free to edit the PR it that helps :-)

lrhn:
Good catch, fixed.

natebosch:
Is this worth revisiting?

Kernald:
> Is this worth revisiting?

As far as I know, either my or @lrhn 's implementations are ready to go - I'm not sure what would be necessary to push it across the line. With that said, I'm personally not using Dart anymore at the moment so I don't have any need for this - it looks like some people found the idea useful from the reactions though.

natebosch:
@lrhn - worth running with your suggested version?

lrhn:
I think this is a bit too specialized and speculative. I can't find a good use for it. That means that any implementation we add risks being too heavily tuned towards the one use-case that we know of, and might not be reusable anyway the next time someone needs something similar.

Should it be in `Iterable` instead of `List`? Or as well? 
Should it return an `Iterable<Iterable<T>>` on iterables?

And when would you want to use it instead of just writing:
```dart
for (var i = 0; i < list.length; i += step) {
  var slice = list.slice(i, min(i + step, list.length));
  ...
}
```

(I'm not a big fan of trying to do all your computations in iterable chains, just because you can. Sometimes it feels more complicated than just doing the loop.)

I'll vote against adding it at this point. Not convinced the problem is ripe for a solution yet.

Kernald:
Closing given the apparent lack of interest.

PR #248, opened 2022-09-28
Update `mergeMaps` making it more flexible

bambinoua:
The updated version of `mergeMaps` function allows to merge maps with removal functionality.
```dart
final map1 = {'a': 'a'};
final map2 = {'b': 'b'};
final map3 = {'c': 'c'};
final map12 = mergeMaps(map1, map2);                     // {'a': 'a', 'b': 'b'}
final map123 = mergeMaps(map12, map3);                   // {'a': 'a', 'b': 'b', 'c': 'c'}
final map123_reduced1 = mergeMaps(map123, {'c': null});  // {'a': 'a', 'b': 'b'}
final map4 = {'a': 'a', 'b': 'b'};
final map123_reduced2 = mergeMaps(map123, map4, 
 removeDifference: true);                                // {'a': 'a', 'b': 'b'}
```

### DIFF
--- a/lib/src/functions.dart
+++ b/lib/src/functions.dart
@@ -30,14 +30,34 @@ Map<K2, V2> mapMap<K1, V1, K2, V2>(Map<K1, V1> map,
 /// select the value that goes into the resulting map based on the two original
 /// values. If [value] is omitted, the value from [map2] is used.
 Map<K, V> mergeMaps<K, V>(Map<K, V> map1, Map<K, V> map2,
-    {V Function(V, V)? value}) {
+    {V? Function(V, V)? value, bool removeDifference = false}) {
   var result = Map<K, V>.of(map1);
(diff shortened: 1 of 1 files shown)


natebosch:
This new behavior seems pretty specialized. I don't think we should have behavior dependent on treating `null` value in some special way - it's not typical to treat present with a null value differently from an absent key, but we give it a special meaning here it will be breaking for anyone that is doing it already.

The `removeDifference` could also be handled outside this method `mergeMaps(aMap, bMap)..removeWhere((k, v) => !bMap.containsKey(k))`. With that behavior it's also feasible to write it with a map literal `{ for (var MapEntry(:key, :value) in bMap.entries) key: aMap.containsKey(key) ? _merge(aMap[key], value) : value}`

I don't think it's worth making `mergeMaps` more complicated for this. cc @lrhn in case you have different preferences.

PR #256, opened 2022-11-23
Tweak docs for split extensions

natebosch:
- Fix the doc for `splitAfter` to use "after" instead of "before".
- Expand the details for the "after" methods with a more explicit
  mention of how the element that satisfies the predicate is the last
  value in it's sublist.
- Use the same input for the example in `splitBefore` as `splitAFter` so
  that the distinction is more clear.

### DIFF
--- a/lib/src/iterable_extensions.dart
+++ b/lib/src/iterable_extensions.dart
@@ -418,7 +418,7 @@ extension IterableExtension<T> on Iterable<T> {
   /// for whether it should be the first element in a new chunk.
   /// If so, the elements since the previous chunk-starting element
   /// are emitted as a list.
-  /// Any final elements are emitted at the end.
+  /// Any remaining elements are emitted at the end.
   ///
   /// Example:
   /// Example:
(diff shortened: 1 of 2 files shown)


lib/src/iterable_extensions.dart
```diff
@@ -415,27 +415,28 @@ extension IterableExtension<T> on Iterable<T> {
   /// Splits the elements into chunks before some elements.
   ///
   /// Each element except the first is checked using [test]
-  /// for whether it should start a new chunk.
+  /// for whether it should be the first element in a new chunk.
   /// If so, the elements since the previous chunk-starting element
   /// are emitted as a list.
   /// Any final elements are emitted at the end.
```
lrhn, line 421: Maybe change "final" to "remaining".

lrhn (approved):

lib/src/iterable_extensions.dart
natebosch: Done

PR #258, opened 2022-12-08
Change `CastError` to `TypeError`, since the former is deprecated.

lrhn:
Does not affect any user code, changes are only in tests and documentation.
Failing casts already throw errors which are both `TypeError` and `CastError`, 
and which will be only `TypeErrors` in Dart 3.0 when `CastError` is removed.


mit-mit (approved):

dcharkes:
Please merge and roll into g3 🙃  It's causing g3 failures on https://dart-review.googlesource.com/c/sdk/+/259041.

mit-mit:
roll: https://dart-review.googlesource.com/c/sdk/+/274740

PR #259, opened 2022-12-21
RFC: Add enumerated extension

lukepighetti:
I have found this extension to be useful, it allows someone to skip the common for loop pattern for accessing elements. It also supersedes the need for specialized methods like `mapIndexed`, `whereIndexed`, (I count 14 in this package). BUT the value seems low because technically we have `asMap` and these specialized methods already

### Before

```dart
for(var i = 0; i < foo.length; i++){
  print(i);
  print(foo[i]);
}
```

```dart
foo.forEach((value){
  final index = foo.indexOf(value);
  print(index);
  print(value);
}
```

### After

```dart
for(final element in foo.enumerated){
  print(element.index);
  print(element.value);
}
```

```dart
foo.enumerated.forEach((element){
  print(element.index);
  print(element.value);
}
```

### Related issues

Lots of JavaScript developers rely on their overloadable map method which allows optionally accessing the index, so tutorials like this are created to shim this missing functionality https://fireship.io/snippets/dart-how-to-get-the-index-on-array-loop-map/

This is a common question on Stack Overflow

Screen Shot 2022-12-21 at 8 49 44 AM

Entire videos on YouTube are dedicated to helping people with this

https://www.youtube.com/watch?v=HLMzibT2Wms

There are some related issues here, although admittedly they are closed

https://github.com/dart-lang/sdk/issues/49589
https://github.com/dart-lang/sdk/issues/33965

### DIFF
--- a/lib/src/iterable_extensions.dart
+++ b/lib/src/iterable_extensions.dart
@@ -591,6 +591,45 @@ extension IterableExtension<T> on Iterable<T> {
       yield slice;
     }
   }
+
+  /// Elements paired with their index.
+  ///
+  /// For example, `['A'].enumerated` returns `[Enumerated('A', 0)]`
+  ///
+  /// Common usage:
+  /// ```dart
+  /// for (var el in ['A', 'B', 'C'].enumerated) {
+  ///   print(el.index); // 0, 1, 2
+  ///   print(el.value); // "A", "B", "C"
+  /// }
+  /// ```
+  Iterable<Enumerated<T>> get enumerated sync* {
+    for (var i = 0; i < length; i++) {
+      yield Enumerated<T>(i, elementAt(i));
+    }
+  }
+}
+
+/// An enumerated element from an iterable
+class Enumerated<T> {
+  /// The index of the element from the source iterable
+  final int index;
+
+  /// The value of the element
+  final T value;
+
+  Enumerated(this.index, this.value);
+
+  @override
+  int get hashCode => index.hashCode ^ value.hashCode;
+
+  @override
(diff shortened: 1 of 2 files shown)


lib/src/iterable_extensions.dart
```diff
@@ -591,6 +591,45 @@ extension IterableExtension<T> on Iterable<T> {
...
+  /// }
+  /// ```
+  Iterable<Enumerated<T>> get enumerated sync* {
+    for (var i = 0; i < length; i++) {
+      yield Enumerated<T>(i, elementAt(i));
+    }
+  }
+}
+
+/// An enumerated element from an iterable
```
natebosch, line 613: We're considering making `MapEntry` an inline class over a `Record`

I wonder if we'd prefer this also be a Record, and if so, if it would be worth waiting for that language feature to publish this API.

cc @lrhn

lukepighetti: I had this same thought, seems like a perfect use case for a Record, and waiting for that to land would make sense to me

lrhn: I'd definitely wait for Dart 3.0 and records and then add (maybe even to the platform libraries):
```dart
extension IterableIndexed<T> on Iterable<T> {
  Iterable<(int, T)> get indexed sync* { 
    var index = 0;
    for (var value in this) yield (index++, value);
  }
}
```
(I've already written it into the platform libraries in one of by branches. No promises that it'll land.)

mosuem:
Closing this as it is stale. Feel free to reopen in case you still want to land it!

PR #261, opened 2022-12-29
`CanonicalizedMap`: new `copy`, `toMap` and `toMapOfCanonicalKeys` methods

gmpassos:
- `copy`: copies a `CanonicalizedMap` instance without recalculating the canonical values of the keys.
- `toMap`: creates a `Map<K,V>` (with the original key values).
- `toMapOfCanonicalKeys`: creates a `Map<C,V>` (with the canocalized keys).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 1.17.3
+## 1.18.0-wip
 
 - `CanonicalizedMap`:
   - Added methods:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+## 1.17.3
+
+- `CanonicalizedMap`:
+  - Added methods:
+    - `copy`: copies an instance without recalculating the canonical values of the keys.
+    - `toMap`: creates a `Map<K,V>` (with the original key values).
+    - `toMapOfCanonicalKeys`: creates a `Map<C,V>` (with the canonicalized keys).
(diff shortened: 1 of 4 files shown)


natebosch:
These APIs look relatively straightforward and I don't have concerns about adding them.

For clarity, @gmpassos - could you describe the use case you have in mind for these?

cc @lrhn

gmpassos:
> It should just be a minor version increment, and a changelog entry for that version saying which methods are added.

Done 👍

CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
+## 1.17.3
```
natebosch, line 1:
```suggestion
## 1.18.0-wip
```

pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: collection
-version: 1.17.2
+version: 1.17.3
```
natebosch, line 2:
```suggestion
version: 1.18.0-wip
```

natebosch (approved):

pubspec.yaml
gmpassos: I can commit the suggestion, but please confirm if you really want to jump to `1.18.0` for this merge.

Best regards.

natebosch: Yes, for new APIs we rev the second number in the semver version.

natebosch (approved):

PR #265, opened 2023-01-17
Changes to Equality classes.

lrhn:
PROOF-OF-CONCEPT. DO NOT LAND.

`MapEquality` and `SetEquality` now expects the provided equalities to be compatible with the inherent equals (and hash) of the map or set.
That allows using `Map.operator[]` and `Set.lookup` to match entries/elements from one collection with those of the other, instead of building an extra
hash-map on the side to count occurrences.

This helps `DeepCollectionEquality` which is likely to traverse a larger collection structure (mainly maps) and do repeated equals/hash code on sup-components.

For unordered `DeepCollectionEquality`, a cache of computed values is added to avoid repeating the same computation during the traversal of a map-structure.

Smaller changes to other classes: `DefaultEquality` and `IdentityEquality` now have only factory constructors, so cannot be extended. There should be no need for that.

Also adds `JsonEquality` which is tuned to JSON-like map/list/basic-value structures, and which assumes everything uses default `==` equality.

NOTICE: This is probably a breaking change, so the version number should be major-version-incremented.
I believe this package is pinned by Flutter, so that'll be fun. :wink:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,31 @@
 
 * Require Dart 2.18
 * Improve docs for `splitAfter` and `splitBefore`.
+* Add `JsonEquality` which is an extra efficient `Equality` on JSON-like
+  object structures. It's less general than `DeepCollectionEquality`,
+  but can therefore make assumptions that allows it to be more efficient.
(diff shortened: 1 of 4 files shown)


mosuem:

PR #267, opened 2023-01-25
add a publish script; prep to publish

devoncarew:
- add a publish script
- prep to publish `1.17.1`

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -1,0 +1,14 @@
+# A CI configuration to auto-publish pub packages.
+
+name: Publish
+
+on:
+  pull_request:
+    branches: [ master ]
+  push:
+    tags: [ 'v[0-9]+.[0-9]+.[0-9]+*' ]
+
+jobs:
+  publish:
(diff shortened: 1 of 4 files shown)


kevmoo (approved):

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,14 @@
+# A CI configuration to auto-publish pub packages.
+
+name: Publish
+
+on:
+  pull_request:
+    branches: [ master ]
```
kevmoo, line 7: why does publish run on a PR?

kevmoo (approved):

.github/workflows/publish.yaml
devoncarew: It validates the info - so you know that it's something you _could_ publish - and it writes a comment to the PR telling you the git tag to use if everything looks good for publishing.

So, validation on PRs, and, publishing on well-formed commits.

kevmoo: Clever clever...

PR #283, opened 2023-05-03
Create version which allows Dart 3.0 SDKs.

lrhn:
This version just allows Dart 3.0 SDKs, a second CL will switch the package to being 3.0 code.
@mit-mit

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -1,10 +1,10 @@
 name: collection
-version: 1.17.1
+version: 1.17.2
 description: Collections and utilities functions and classes related to collections.
 repository: https://github.com/dart-lang/collection
 
 environment:
-  sdk: ">=2.18.0 <3.0.0"
+  sdk: ">=2.18.0 <4.0.0"
 
 dev_dependencies:
   lints: ^2.0.0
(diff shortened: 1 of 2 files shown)


pubspec.yaml
```diff
@@ -1,10 +1,10 @@
 name: collection
-version: 1.17.1
+version: 1.17.2
 description: Collections and utilities functions and classes related to collections.
 repository: https://github.com/dart-lang/collection
 
 environment:
-  sdk: ">=2.18.0 <3.0.0"
+  sdk: ">=2.18.0 <4.0.0"
```
mit-mit, line 7: This really makes no difference, as we re-write the upper constraint to `4.0.0` anyway for null safe packages.

But don't we want to change the lower constraint too as you are about to start using Dart 3 modifiers?

lrhn: I prefer to not *depend* on constraint re-writing. It's a hack. It's fine transitionally, but I'd like to have a version which opts in to `<4.0.0`.

But I already had a PR for that, so closing this one.

lrhn:
Whoops, I already had a CL for this.

PR #284, opened 2023-05-03
Update code to Dart 3.0

lrhn:
Add required `mixin`s. Add some appropriate `final`s.

Should this be a major version increment?

The `final` modifiers are breaking. (But so is the lack of `mixin` modifiers on some classes.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 * Require Dart 3.0.
 * Mark mixin classes as `mixin class`.
+* Added a `Map.pairs` extension method to make it easier to work with maps using
+  Dart 3 record types.
 
 ## 1.17.2
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 1.18.0
+
+* Require Dart 3.0.
+* Mark mixin classes as `mixin class`.
+
 ## 1.17.2
 
 * Accept Dart SDK versions above 3.0.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
+## 1.17.2
+
+* Accept Dart SDK versions above 3.0.
+
 ## 1.17.1
 
-* Require Dart 2.18
+* Require Dart 2.18.
 * Improve docs for `splitAfter` and `splitBefore`.
 
 ## 1.17.0
--- a/lib/src/canonicalized_map.dart
+++ b/lib/src/canonicalized_map.dart
@@ -10,7 +10,7 @@ import 'dart:collection';
 /// more efficient than a [LinkedHashMap] with a custom equality operator
 /// because it only canonicalizes each key once, rather than doing so for each
 /// comparison.
-class CanonicalizedMap<C, K, V> implements Map<K, V> {
+final class CanonicalizedMap<C, K, V> implements Map<K, V> {
   final C Function(K) _canonicalize;
 
   final bool Function(K)? _isValidKeyFn;
--- a/lib/src/equality.dart
+++ b/lib/src/equality.dart
@@ -9,7 +9,7 @@ import 'comparators.dart';
 const int _hashMask = 0x7fffffff;
 
 /// A generic equality relation on objects.
-abstract class Equality<E> {
+abstract interface class Equality<E> {
   const factory Equality() = DefaultEquality<E>;
 
   /// Compare two elements for being equal.
@@ -46,7 +46,7 @@ abstract class Equality<E> {
 ///
 /// It's also possible to pass an additional equality instance that should be
 /// used to compare the value itself.
-class EqualityBy<E, F> implements Equality<E> {
+final class EqualityBy<E, F> implements Equality<E> {
   final F Function(E) _comparisonKey;
 
   final Equality<F> _inner;
@@ -81,7 +81,7 @@ class EqualityBy<E, F> implements Equality<E> {
 /// Note that [equals] and [hash] take `Object`s rather than `E`s. This allows
 /// `E` to be inferred as `Null` in const contexts where `E` wouldn't be a
 /// compile-time constant, while still allowing the class to be used at runtime.
-class DefaultEquality<E> implements Equality<E> {
+final class DefaultEquality<E> implements Equality<E> {
   const DefaultEquality();
   @override
   bool equals(Object? e1, Object? e2) => e1 == e2;
@@ -92,7 +92,7 @@ class DefaultEquality<E> implements Equality<E> {
 }
 
 /// Equality of objects that compares only the identity of the objects.
-class IdentityEquality<E> implements Equality<E> {
+final class IdentityEquality<E> implements Equality<E> {
   const IdentityEquality();
   @override
   bool equals(E e1, E e2) => identical(e1, e2);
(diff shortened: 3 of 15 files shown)


pubspec.yaml
```diff
@@ -1,10 +1,10 @@
 name: collection
-version: 1.17.2
+version: 1.18.0
 description: Collections and utilities functions and classes related to collections.
 repository: https://github.com/dart-lang/collection
 
 environment:
-  sdk: ">=2.18.0 <4.0.0"
+  sdk: ">=3.0.0 <4.0.0"
```
devoncarew, line 7: Or just `^3.0.0`?

natebosch:
> Should this be a major version increment?

For most packages, yes.

For this package, which is pinned by flutter, it may be easier to cheat and ship without a major version change.

lrhn:
I don't think we're time-bound on this release. The internal need to make the code 3.0 compatible has been fixed, so staying as 2.18.0 is fine for now.
We want to release a 3.0 version eventually, where we add modifiers (and the sooner we do so, the sooner people can see which modifiers they need to comply with).

pubspec.yaml
lrhn: It might be cargo-cult at this point, but at some point there were some tools which couldn't handle `^...` notation in SDK versions.
Are we sure those tools are gone?

lrhn:
(Need to fix CI too.)

pubspec.yaml
natebosch: Yes, the server filters these for SDK versions that can't read them. It should be safe to use caret constraints for packages that only support Dart 3

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 1.18.0
+
+* Require Dart 3.0.
+* Mark mixin classes as `mixin class`.
+
 ## 1.17.2
```
natebosch, line 6: This version was never published, so we can probably just drop the changelog entry.

natebosch:
I'm starting an internal presubmit to see if any usage is impacted by the change.

lib/src/unmodifiable_wrappers.dart
```diff
@@ -25,7 +25,7 @@ class NonGrowableListView<E> extends DelegatingList<E>
 
 /// Mixin class that implements a throwing version of all list operations that
 /// change the List's length.
-abstract class NonGrowableListMixin<E> implements List<E> {
+abstract mixin class NonGrowableListMixin<E> implements List<E> {
```
natebosch, line 28: Also need to remove comments around `mixin` on line 167 for `UnmodifiableMapMixin`

lib/src/canonicalized_map.dart
```diff
@@ -10,7 +10,7 @@ import 'dart:collection';
 /// more efficient than a [LinkedHashMap] with a custom equality operator
 /// because it only canonicalizes each key once, rather than doing so for each
 /// comparison.
-class CanonicalizedMap<C, K, V> implements Map<K, V> {
+final class CanonicalizedMap<C, K, V> implements Map<K, V> {
```
natebosch, line 13: This class has subclasses in the ecosystem.

https://github.com/dart-lang/http_parser/blob/bbe37dd228ec59f58a73df4b328ef747757165c7/lib/src/case_insensitive_map.dart#L10

I think this looks like a legitimately useful pattern. Perhaps this class could be `base` instead of `final`?

natebosch: I guess even marking it as `base` will be breaking for this use case... not sure if we want to leave it loose, or ignore the fact that it is breaking (and submit a fix at the usage site ahead of time)

pubspec.yaml
kevmoo: @lrhn – it was like Dart 1.8.

lib/src/equality.dart
```diff
@@ -418,7 +419,7 @@ class MultiEquality<E> implements Equality<E> {
 ///
 /// A list is only equal to another list, likewise for sets and maps. All other
 /// iterables are compared as iterables only.
-class DeepCollectionEquality implements Equality {
+final class DeepCollectionEquality implements Equality {
```
natebosch, line 422: Mockito extends this class.

https://github.com/dart-lang/mockito/blob/56173fa356c7017e1d2137ce3600963518bf497d/lib/src/invocation_matcher.dart#L155

We could rewrite that to be `implements Equality` and use a `const DeepCollectionEquality()` to reuse the required behavior - I might take a look at doing that anyway.

lib/src/equality.dart
```diff
@@ -81,7 +81,7 @@ class EqualityBy<E, F> implements Equality<E> {
 /// Note that [equals] and [hash] take `Object`s rather than `E`s. This allows
 /// `E` to be inferred as `Null` in const contexts where `E` wouldn't be a
 /// compile-time constant, while still allowing the class to be used at runtime.
-class DefaultEquality<E> implements Equality<E> {
+final class DefaultEquality<E> implements Equality<E> {
```
natebosch, line 84: https://github.com/angulardart/angular/blob/e943f6f0b8f8ae0d64c2c7935d4554f41d86e2cd/angular/lib/src/runtime/check_binding.dart#L186-L190


This has a usage of `extends`

pubspec.yaml
lrhn: ACK, good to know!

lib/src/canonicalized_map.dart
natebosch: > Why is making it `base` breaking?

It could be a quirk of google3 - when I made it `base` I saw a build failure until I also marked the subclass as `base`.



> In either case, the result doesn't need to be-a `CanonicalizedMap`. That's leaking an implementation detail.

Yeah, it's not a good design and we should fix it. We just need to decide what types of misuses that exist in google3 or that we can find in pub that we want to avoid breaking, and which we are OK with breaking.

natebosch:
The full set of modifiers in this CL which caused build failures in google3 are `CanonicalizedMap`, `DefaultEquality`, `MapEquality`, `DeepCollectionEquality`, and `HeapPriorityQueue`. I think you fixed the usage of `HeapPriorityQueue` but I think the others will still cause a problem. When I comment out the added modifiers on those classes I can get a TAP global without any related build failures.

lib/src/map_extensions.dart
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+extension MapExtensions<K, V> on Map<K, V> {
+  /// Like [Map.entries], but returns each entry as a record.
+  Iterable<(K, V)> get pairs => entries.map((e) => (e.key, e.value));
```
kevmoo, line 7: Interesting...but why? What's the upside here? What do users get?

We're already going to allocate (however short-lived) `MapEnty` instances.

natebosch: The main benefit is being able to loop over pairs with variables named something other than "key" and "value".

https://github.com/dart-lang/collection/pull/289#issuecomment-1559974806

natebosch: I think we're still hoping to make `MapEntry` an inline class over a `(K, V)`, and if we are able to do that this `.pairs` extension will be useless. I think it's worth having it early in this package.

lrhn: I doubt it will be over `(K, V)`. More likely it's `({K key, V value})`, since that will allow `dynamic`-typed code to keep working.

nex3:
What's the status of this PR? I'm itching to use `Map.pairs` in my packages.

dannnnthemannnn:
Hey all, I'm trying to use collections.dart and am getting errors like:

```
packages/collection/src/boollist.dart:231:50:
Error: The class 'NonGrowableListMixin' can't be used as a mixin because it isn't a mixin class nor a mixin.
class _NonGrowableBoolList extends BoolList with NonGrowableListMixin<bool> {
```

This is pulled in via transitive dependencies. 

My dart version:
```
dart --version
Dart SDK version: 3.3.1 (stable) (Wed Mar 6 13:09:19 2024 +0000) on "macos_arm64"
```

I see above people saying it isnt a problem but how can I mitigate this?

mosuem:

PR #297, opened 2023-07-13
Make sure .sample()'s result is randomized in more cases

ebraminio:
Sampling result was always not randomized when requested samples count was equal or smaller than the input list length and was lots of times not randomized when requested samples count was a little smaller than input list length.

This fixes it simply by using core provided .shuffle() and emulates sampling on top of it and provides tests that were failing on the older code but now don't fail.

The older version was only useful when the requested samples count was significantly smaller than total length of the list which isn't always the case and acts surprisingly on the mentioned cases and I'd believe we should we should follow principle of least astonishment as this has made some issues for my flutter app and I've brought an abstract reproduction of the issue below,

Before this change:
```dart
final random = Random(123);
print([1, 2, 3, 4, 5].sample(1, random)); // [3] a random pick
print([1, 2, 3, 4, 5].sample(2, random)); // [5, 4] two random picks
print([1, 2, 3, 4, 5].sample(3, random)); // [1, 2, 5] three random pick, not randomized enough
print([1, 2, 3, 4, 5].sample(4, random)); // [1, 2, 3, 4] at this specific length and random seed no randomization has happened
print([1, 2, 3, 4, 5].sample(5, random)); // [1, 2, 3, 4, 5] no matter the seed no randomization will happen
print([1, 2, 3, 4, 5].sample(6, random)); // [1, 2, 3, 4, 5] //
print([1, 2, 3, 4, 5].sample(10, random)); // [1, 2, 3, 4, 5] //
```

Some of the entries aren't randomized enough. With this change however:
```dart
final random = Random(123);
print([1, 2, 3, 4, 5].sample(1, random)); // [3] randomized pick
print([1, 2, 3, 4, 5].sample(2, random)); // [4, 5] randomized picks
print([1, 2, 3, 4, 5].sample(3, random)); // [5, 1, 4] //
print([1, 2, 3, 4, 5].sample(4, random)); // [1, 2, 4, 3] //
print([1, 2, 3, 4, 5].sample(5, random)); // [5, 3, 2, 1, 4] //
print([1, 2, 3, 4, 5].sample(6, random)); // [5, 2, 1, 3, 4] //
print([1, 2, 3, 4, 5].sample(10, random)); // [4, 3, 2, 5, 1] //
```

### DIFF
--- a/lib/src/iterable_extensions.dart
+++ b/lib/src/iterable_extensions.dart
@@ -30,23 +30,8 @@ extension IterableExtension<T> on Iterable<T> {
   /// The chosen elements are not in any specific order.
   List<T> sample(int count, [Random? random]) {
     RangeError.checkNotNegative(count, 'count');
-    var iterator = this.iterator;
-    var chosen = <T>[];
-    for (var i = 0; i < count; i++) {
-      if (iterator.moveNext()) {
-        chosen.add(iterator.current);
-      } else {
-        return chosen;
-      }
-    }
-    var index = count;
-    random ??= Random();
-    while (iterator.moveNext()) {
-      index++;
-      var position = random.nextInt(index);
-      if (position < count) chosen[position] = iterator.current;
-    }
-    return chosen;
+    var shuffled = toList()..shuffle(random ??= Random());
+    return shuffled.take(count).toList();
   }
 
   /// The elements that do not satisfy [test].
--- a/lib/src/iterable_extensions.dart
+++ b/lib/src/iterable_extensions.dart
@@ -30,8 +30,8 @@ extension IterableExtension<T> on Iterable<T> {
   /// The chosen elements are not in any specific order.
   List<T> sample(int count, [Random? random]) {
     RangeError.checkNotNegative(count, 'count');
-    var shuffled = toList()..shuffle(random ??= Random());
-    return shuffled.take(count).toList();
+    var shuffled = [...this]..shuffle(random ??= Random());
+    return [...shuffled.take(count)];
   }
 
   /// The elements that do not satisfy [test].
--- a/lib/src/iterable_extensions.dart
+++ b/lib/src/iterable_extensions.dart
@@ -30,7 +30,7 @@ extension IterableExtension<T> on Iterable<T> {
   /// The chosen elements are not in any specific order.
   List<T> sample(int count, [Random? random]) {
     RangeError.checkNotNegative(count, 'count');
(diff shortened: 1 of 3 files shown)


natebosch:
The comment mentions `The chosen elements are not in any specific order.` so it does not make any promise that the result is shuffled, and retaining the input order seems valid to me. I don't think it's a good idea to try to allocate the full list to shuffle - the current impl would work for very long iterables without using more memory than required.

I don't think we should land this PR.

natebosch:
If we do decide to return a shuffled result (cc @lrhn in case this is something you do want to do - I can see _some_ value in shuffling, just to enforce that the ordering cannot be relied on) then I think we'd want

```diff
diff --git a/lib/src/iterable_extensions.dart b/lib/src/iterable_extensions.dart
index 7772622..55df89f 100644
--- a/lib/src/iterable_extensions.dart
+++ b/lib/src/iterable_extensions.dart
@@ -46,7 +46,7 @@ extension IterableExtension<T> on Iterable<T> {
       var position = random.nextInt(index);
       if (position < count) chosen[position] = iterator.current;
     }
-    return chosen;
+    return chosen..shuffle();
   }
 
   /// The elements that do not satisfy [test].
```

This way it wouldn't need to allocate a list longer than `count`. I'm going to close the PR for now, and we can go forward with that approach if we do decide to shuffle the result.

lib/src/iterable_extensions.dart
```diff
@@ -30,23 +30,8 @@ extension IterableExtension<T> on Iterable<T> {
...
-    }
-    var index = count;
-    random ??= Random();
-    while (iterator.moveNext()) {
-      index++;
-      var position = random.nextInt(index);
-      if (position < count) chosen[position] = iterator.current;
-    }
-    return chosen;
+    var shuffled = [...this]..shuffle(random ?? Random());
```
lrhn, line 33: This implementation requires keeping all the iterable's elements in memory at the same time.
If you do that anyway, the `sample` method doesn't make much sense.
The benefit of `sample` is that you can get *n* elements without using more than O(*n*) space, which makes sense for very long iterables.

lrhn:
We can shuffle at the end, or after collecting the initial list. Either should work. At the end is probably simpler.

ebraminio:
> We can shuffle at the end, or after collecting the initial list. Either should work. And at the early return too.

Ok then, https://github.com/dart-lang/collection/pull/302

PR #298, opened 2023-07-13
Add IterableExtension.shuffled

ebraminio:
Creates a shuffled list of the elements of the iterable.

For example Kotlin has an immutable `shuffled` https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/shuffled.html in addition a mutable one so thought might be useful here also as wanted to use it in my other PR.


natebosch (approved):
Seems useful enough to me. Any concerns @lrhn ?

lrhn:
LGTM. Consistent with `sorted`.

ebraminio:
Just rebased it, thanks

PR #300, opened 2023-07-26
add onEach for Iterable

HuixingWong:
Just add onEach function


HuixingWong:
Seems no one need this, I'll just use it myself

PR #302, opened 2023-08-10
Make sure .sample()'s result is randomized in more cases

ebraminio:
Sampling result was always not randomized when requested samples count was equal or smaller than the input list length and was lots of times not randomized when requested samples count was a little smaller than input list length.

This fixes it simply by using core provided .shuffle() and provides tests that was failing on the older code but now don't fail.

As https://github.com/dart-lang/collection/pull/297#issuecomment-1673153583

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - Fixes bugs in `ListSlice.slice` and `ListExtensions.slice`.
 - Adds `shuffled` to `IterableExtension`.
 - Update to `package:lints` 2.0.1.
+- Better randomization of `IterableExtension.sample` results.
 
 ## 1.17.2
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,8 @@
     - `toMapOfCanonicalKeys`: creates a `Map<C,V>` (with the canonicalized keys).
 - Fixes bugs in `ListSlice.slice` and `ListExtensions.slice`.
 - Adds `shuffled` to `IterableExtension`.
+- Shuffle `IterableExtension.sample` results.
 - Update to `package:lints` 2.0.1.
-- Better randomization of `IterableExtension.sample` results.
 
 ## 1.17.2
 
--- a/lib/src/iterable_extensions.dart
+++ b/lib/src/iterable_extensions.dart
@@ -36,7 +36,7 @@ extension IterableExtension<T> on Iterable<T> {
       if (iterator.moveNext()) {
         chosen.add(iterator.current);
       } else {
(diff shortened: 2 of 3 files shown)


lrhn (approved):
LGTM now. @natebosch ?

lrhn:
(I don't know *why* randomness is needed, we never promised that the ordering was random, only the selection, but if it makes people less likely to depend on a specific ordering, or less likely to get issues by *assuming* that the ordering is random just because it looks like that for large number, then it's probably a good thing. But it is an overhead that isn't really needed.)

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
     - `toMapOfCanonicalKeys`: creates a `Map<C,V>` (with the canonicalized keys).
 - Fixes bugs in `ListSlice.slice` and `ListExtensions.slice`.
 - Update to `package:lints` 2.0.1.
+- Better randomization of `IterableExtension.sample` results.
```
natebosch, line 10:
```suggestion
- Shuffle `IterableExtension.sample` results.
```

natebosch (approved):

lib/src/iterable_extensions.dart
```diff
@@ -36,7 +36,7 @@ extension IterableExtension<T> on Iterable<T> {
       if (iterator.moveNext()) {
         chosen.add(iterator.current);
```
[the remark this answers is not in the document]
ebraminio: the new code (after replacing `for` with `while` in it) fails at this part of the test suit https://github.com/ebraminio/dart-collection/blob/randomized-sample-2/test/extensions_test.dart#L1162-L1165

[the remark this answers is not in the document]
ebraminio: Excellent solution, done

lib/src/iterable_extensions.dart
```diff
@@ -32,15 +32,22 @@ extension IterableExtension<T> on Iterable<T> {
...
       if (iterator.moveNext()) {
-        chosen.add(iterator.current);
+        var nextElement = iterator.current;
+        var position = random.nextInt(chosen.length + 1);
+        if (position == chosen.length) {
+          chosen.add(nextElement);
+        } else {
+          chosen.add(chosen[position]);
+          chosen[position] = nextElement;
+        }
```
natebosch, line 45:
```suggestion
        chosen.insert(position, nextElement);
```

natebosch: Hmm, maybe this could be slower...

natebosch (approved):

PR #305, opened 2023-08-15
Add noneIndexed, everyIndexed and anyIndexed

ebraminio:
Adds indexed counterpart of none, every and any.
    
I had to remove none, every and any when I needed the element index also so this adds indexed counterpart of them.

JavaScript's every and some also provide index,

 `[1, 2, 3].every((x, i) => (console.log(x, i), true))`

image

`['a', 'b', 'c'].some(console.log)`

image

### DIFF
--- a/lib/src/iterable_extensions.dart
+++ b/lib/src/iterable_extensions.dart
@@ -582,6 +582,59 @@ extension IterableExtension<T> on Iterable<T> {
     return true;
   }
 
+  /// Whether no element and index satisfies [test].
+  ///
+  /// Returns true if no element and index satisfies [test],
+  /// and false if at least one does.
+  ///
+  /// Equivalent to `iterable.everyIndexed((i, x) => !test(i, x))` or
+  /// `!iterable.anyIndexed(test)`.
+  bool noneIndexed(bool Function(int, T) test) {
+    var index = 0;
+    for (var element in this) {
+      if (test(index++, element)) return false;
+    }
+    return true;
+  }
+
+  /// Checks whether every element and index of this iterable satisfies [test].
+  ///
+  /// Checks every element and index in iteration order, and returns `false` if
+  /// any of them make [test] return `false`, otherwise returns `true`.
+  ///
+  /// Example:
+  /// ```dart
+  /// final elements = [0, 1, 2];
+  /// // Checks whether every element index matches its value.
+  /// final every = elements.everyIndexed((index, value) => index == value); // true
+  /// ```
+  bool everyIndexed(bool Function(int, T) test) {
+    var index = 0;
+    for (var element in this) {
+      if (!test(index++, element)) return false;
+    }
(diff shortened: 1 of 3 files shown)


natebosch (approved):
LGTM. Any concerns @lrhn ?

lib/src/iterable_extensions.dart
```diff
@@ -582,6 +582,59 @@ extension IterableExtension<T> on Iterable<T> {
...
+  ///
+  /// Equivalent to `iterable.everyIndexed((i, x) => !test(i, x))` or
+  /// `!iterable.anyIndexed(test)`.
+  bool noneIndexed(bool Function(int, T) test) {
+    var index = 0;
+    for (var element in this) {
+      if (test(index++, element)) return false;
+    }
+    return true;
+  }
```
lrhn, line 598: I'd use `iterable.indexed.none((p) => test(p.$1, p.$2))` and wait for pattern parameters to make that just 
```dart
iterable.indexed.none((var (index, element)) { 
  // test here
});
```
I don't think we should be adding methods that introduce indexes, instead we should make things work with the `.indexed`  we have in the platform libraries, which works on all iterables.

natebosch: That's a good point - `indexed` is pretty composable with most of the operations other than the various `*WhereIndexed*`. We might even consider deprecating them and add a fix to use the `.indexed.operation` variant.

ebraminio:
Use of `indexed` can make sense to me also, so let's close this.

lib/src/iterable_extensions.dart
ebraminio: > [..] wait for pattern parameters to make that just
> 
> ```dart
> iterable.indexed.every((var (index, element)) { 
>   // test here, negated.
> });
> ```

This doesn't seem to work, is it planed for future or the syntax is different?

image

ebraminio: Unrelated to the topic but spotted a compiler crash with it also btw https://github.com/dart-lang/sdk/issues/53268

ebraminio: I removed all uses of the .thingsIndexed in my app which made things more fun (as I can use comprehension also better), just that I wished I had access to the mentioned destructing in argument syntax which I guess in future.

natebosch: > is it planed for future or the syntax is different?

It's something we're considering for the future, but we do not have concrete plans. https://github.com/dart-lang/language/issues/3001

ebraminio: > > is it planed for future or the syntax is different?
> 
> It's something we're considering for the future, but we do not have concrete plans. [dart-lang/language#3001](https://github.com/dart-lang/language/issues/3001)

It's totally understandable 🙏

PR #307, opened 2023-08-20
Add Iterable.countOccurrences

ebraminio:
As a part https://github.com/dart-lang/collection/pull/306#issuecomment-1682281034

Just am asking again whether you might like creating a [dedicated Counter class just like python](https://docs.python.org/3/library/collections.html#collections.Counter) or a `Map<T, int>` is fine, at least temporarily, I would be happy with either way (and that's why I have considered to create another PR for `removeCounted` after the decision for this)


mosuem:

PR #309, opened 2023-08-20
Add Iterable.intersperse extension method

ebraminio:
I needed something that could add an element between each two elements in a list as I just hardcoded that functionality in my flutter app but now have decided to revise that code so did some research about it as I didn't even know what name that has and eventually reached to https://github.com/apple/swift-algorithms/blob/main/Guides/Intersperse.md which itself is linked to https://hackage.haskell.org/package/base-4.14.0.0/docs/Data-List.html#v:intersperse and https://docs.rs/itertools/0.9.0/itertools/trait.Itertools.html#method.intersperse which felt that makes it enough popular to be added to `collection`. It would be ok also if you would like to rename this to something better.

Or maybe we should have two of them, `intersperse` that mutates a list and `interspersed`, the one who returns a new iterator, or maybe we should provide only one and should be named `interspersed` like swift or we should follow rust and Haskell instead.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## 1.19.0-wip
 
-- Adds `shuffled` to `IterableExtension`.
+- Add `Iterable.shuffled` and `Iterable.intersperse` extension methods.
 - Shuffle `IterableExtension.sample` results.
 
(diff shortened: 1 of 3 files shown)


mosuem:

PR #312, opened 2023-09-06
Add ListExtensions.firstOrNull, lastOrNull, singleOrNull.

scheglov:


natebosch:
@lrhn - should we define these in the SDK instead?

scheglov:
This is not urgent at all, but it would be nice to get some reaction to it :-)

natebosch:
I like the APIs. I think it would be worth moving to the SDK since that is where the definitions are for `Iterable`. Unless @lrhn has concerns I'd be inclined to close this and reopen as a CL in the SDK.

lrhn:
SGTM, move it!

natebosch:
@scheglov - do you want to open a CL in the SDK or should I?

scheglov:
I opened https://github.com/dart-lang/sdk/issues/53756

PR #315, opened 2023-10-10
docs: Add a warning about dart-lang/sdk#43763 to sortedBy

### LINKED ISSUE #649: IterableExtension sortedBy doesn't permit sorting by int without workaround
```dart
import 'package:collection/collection.dart';

void main() {
  [2, 1].sortedBy((i) => i);
}
```
yields
```
Error: Inferred type argument 'int' doesn't conform to the bound
'Comparable<K>' of the type variable 'K' on 'IterableExtension|sortedBy'.
```
sorting by an `int` is a pretty important use case.
the problem seems to be that `sortedBy` insists that its generic param `K` extend `Comparable<K>`
and `int` doesn't extend `Comparable<int>`.  rather, it extends `Comparable<num>`.

workaround:
```dart
  [2, 1].sortedBy<num>((i) => i);
```

the workaround doesn't seem like it should be the end of this story.
i feel like the first example should just work.   perhaps one could see this as a problem with the type inference system,
as in, why can't it figure out that `K` perhaps should be bound to a super class (`num`) of the exact type (`int`)
returned by the `keyOf` function?

eseidel:
dart-lang/sdk#43763 is supposedly hard to fix in the language, but is a common gotcha
when dealing with sortedBy, so the least we can do is at least alert users and tell them
how to work around it.

Fixes https://github.com/dart-lang/core/issues/649.

---

- [X] I’ve reviewed the contributor guide and applied the relevant portions to this PR.

<details>

- Most changes should add an entry to the changelog and may need to [rev the pubspec package version](https://github.com/dart-lang/sdk/wiki/External-Package-Maintenance#making-a-change).

</details>

### DIFF
--- a/lib/src/iterable_extensions.dart
+++ b/lib/src/iterable_extensions.dart
@@ -72,6 +72,22 @@ extension IterableExtension<T> on Iterable<T> {
   ///
   /// The elements are ordered by the natural ordering of the
   /// property [keyOf] of the element.
+  ///
+  /// Due to https://github.com/dart-lang/sdk/issues/43763
+  /// if you are comparing using int, you will need to use the <num> type
+  /// annotation instead, e.g.
+  ///
+  /// ```dart
+  /// class Foo {
+  ///  final int bar;
+  ///  Foo(this.bar);
+  /// }
+  ///
+  /// final foos = [Foo(1), Foo(2), Foo(3)];
+  /// // final sortedFoos = foos.sortedBy((foo) => foo.bar); // Error
+  /// final sortedFoos = foos.sortedBy((foo) => foo.bar as num); // Works
+  /// final alsoWorks = foos.sortedBy<num>((foo) => foo.bar); // Works
+  /// ```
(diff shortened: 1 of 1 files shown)


devoncarew (commented):
Thanks for the contribution! Can you add an entry in the changelog for this?

@natebosch or @lrhn - can you review?

natebosch:
I would prefer if this was handled in a more centralized way. https://github.com/dart-lang/sdk/issues/43763#issuecomment-1756007200

If we can't accomplish this in the analyzer we could work on the phrasing here.

eseidel:
I'm not at all wedded to this solution.  Just witnessed myself and two other colleagues each independently (!) hit this error in the last few months and so went to document where it might be seen by others.  I wasn't aware of https://github.com/dart-lang/sdk/issues/43763 until this morning (and managed to fumble my way to a solution in my code without finding it), but after reading https://github.com/dart-lang/sdk/issues/43763 and seeing it's supposedly a 10 year old issue with no easy fix, felt like documenting the failure to avoid others being confused by it could at least help. 🤷

mosuem:

PR #322, opened 2023-11-08
Latest lints, require Dart 3.1, use mixin

kevmoo:


lib/src/empty_unmodifiable_set.dart
```diff
@@ -34,7 +34,7 @@ class EmptyUnmodifiableSet<E> extends IterableBase<E>
   E singleWhere(bool Function(E) test, {E Function()? orElse}) =>
       orElse != null ? orElse() : throw StateError('No element');
   @override
-  Iterable<T> whereType<T>() => Iterable.empty();
+  Iterable<T> whereType<T>() => const Iterable.empty();
```
lrhn, line 37: This is not semantics preserving. The original returned an `Iterable<T>.empty()`, this returns an `Iterable<Never>.empty()`.

It matters if someone does `.toList().add(aTValue)`.
(It probably doesn't, but it *could*.)

So add the `<T>` instead of the `const`.

test/algorithms_test.dart
```diff
@@ -389,10 +390,9 @@ class OC implements Comparable<OC> {
   String toString() => 'OC[$id,$order]';
 }
 
-int ocId(OC oc) => oc.id;
-int ocOrder(OC oc) => oc.order;
+int _ocOrder(OC oc) => oc.order;
 
-int compareInt(int a, int b) => a - b;
+int _compareInt(int a, int b) => a - b;
```
lrhn, line 395: (Test files should *not* have to care whether their functions are public or not. They likely shouldn't be. But if it's just for consistency, that's fine.)

lrhn (approved):
One bug to fix, otherwise OK.

test/algorithms_test.dart
kevmoo: I like consistency. It makes parsing the file easier. Is this used outside the library or not, etc...

lib/src/empty_unmodifiable_set.dart
kevmoo: Done!

PR #328, opened 2024-01-12
Add flattenedToList and flattenedToSet

natebosch:
For iterables which are known to be exhausted after flattening
performance is better than the `sync*` implementation using a collection
literal.

Add `flattenedToList` as a performance improvement over `flattened.`
Add `flattenedToSet` as new behavior for flattening to unique elements.

Originally implemented in
https://github.com/dart-lang/sdk/commit/8d3b6ce54ccb09dd970cd45fc3cded4a35beac31

### DIFF
--- a/test/extensions_test.dart
+++ b/test/extensions_test.dart
@@ -1034,6 +1034,56 @@ void main() {
               [1, 2, 3, 4]);
         });
       });
+      group('.flattenedToList', () {
+        var empty = iterable(<int>[]);
+        test('empty', () {
+          expect(iterable(<Iterable<int>>[]).flattenedToList, []);
+        });
+        test('multiple empty', () {
+          expect(iterable([empty, empty, empty]).flattenedToList, []);
+        });
+        test('single value', () {
+          expect(
+              iterable(<Iterable>[
+                iterable([1])
+              ]).flattenedToList,
+              [1]);
+        });
+        test('multiple', () {
+          expect(
+              iterable(<Iterable>[
+                iterable([1, 2]),
+                empty,
+                iterable([3, 4])
+              ]).flattenedToList,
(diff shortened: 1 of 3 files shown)


natebosch:
cc @scheglov - any concerns about landing this here? You might need to handle import conflicts when this version gets pulled in to the SDK.

natebosch:
I'm not sure if we want to hold this on https://github.com/dart-lang/collection/pull/284 - we also have another new API on that branch waiting for publish.

test/extensions_test.dart
```diff
@@ -1034,6 +1034,56 @@ void main() {
...
+              {1});
+        });
+        test('multiple', () {
+          expect(
+              iterable(<Iterable>[
+                iterable([1, 2]),
+                empty,
+                iterable([3, 4])
+              ]).flattenedToSet,
+              {1, 2, 3, 4});
```
lrhn, line 1084: Also test that duplicates are removed.

lrhn (approved):

lrhn:
If someone needs this, it's fine to release it in 2.X, and merge it into the 3.0 branch too.
Otherwise it's probably easier to just put it into the next major release.

I really should get back to releasing the 3.0 version.

scheglov:
> cc @scheglov - any concerns about landing this here? You might need to handle import conflicts when this version gets pulled in to the SDK.

https://dart-review.googlesource.com/c/sdk/+/346163 renames these methods in the analyzer, so there will be no conflict. I will migrate once this PR lands, and `package:collection` is pulled into SDK.

kevmoo:
@natebosch – care to proceed?

test/extensions_test.dart
kevmoo: done!

PR #333, opened 2024-02-01
Test on dart2wasm

kevmoo:


natebosch:
@jakemac53 - we don't expect significant changes to wasm support in the test runner do we? Any concerns if we start using the current implementation on CI?

kevmoo:
I've already wired it up for a few other packages, FYI – with @jakemac53 approval

natebosch (approved):

jakemac53:
Yeah, it isn't under any experiment flags any longer, we should be good to use it wherever.

PR #335, opened 2024-02-03
Add transitive reduction to the functions library

osaxma:
This PR adds [transitive reduction](https://en.wikipedia.org/wiki/Transitive_reduction) to the functions library. 

I copied and then modified the existing `transitiveClosure` function to reduce the transitive connections instead of flattening them. Similarly for tests, they are the same as the transitive closure ones, but the input and expected results were swapped (where appropriate). 

---

<details>

- Most changes should add an entry to the changelog and may need to [rev the pubspec package version](https://github.com/dart-lang/sdk/wiki/External-Package-Maintenance#making-a-change).

</details>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
   generic.
 - Require Dart `^3.1.0`
 - Mark "mixin" classes as `mixin`.
+- Add `transitiveReduction` to the functions library.
 
 ## 1.18.0
 
--- a/lib/src/functions.dart
+++ b/lib/src/functions.dart
@@ -147,6 +147,43 @@ Map<T, Set<T>> transitiveClosure<T>(Map<T, Iterable<T>> graph) {
   return result;
 }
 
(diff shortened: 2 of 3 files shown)


osaxma:
I think I rushed in submitting this as it doesn't handle the simple case where a path has more than two edges (added a failing test for that). So I will convert it to a draft. 

For reference, it seems the following algorithm, which uses path matrix, is popular:

> Harry T. Hsu. 1975. An Algorithm for Finding a Minimal Equivalent Graph of a Digraph. J. ACM 22, 1 (Jan. 1975), 11–16. https://doi.org/10.1145/321864.321866

osaxma:
In case anyone gets here through search, here is an implementation that works with `List<SomeNode>`. I don't feel it's a good fit for the functions library since it does not work with cycles and it doesn't take `Map<T, Iterable<T>>` as graph similar to the rest of the graph functions. 

- https://gist.github.com/osaxma/9565b39e90785689ddbc7603e1fc3119

PR #336, opened 2024-02-13
Deprecate transitiveClosure

natebosch:
The algorithm is implemented with a more flexible signature in
`package:graphs`.
See https://github.com/dart-lang/tools/issues/104

Use an `ignore_for_file` rather than comment each usage of the
deprecated APIs.


kevmoo:
Changelog?

kevmoo (approved):

PR #343, opened 2024-05-20
`separated` method added in Iterable extension

### LINKED ISSUE #342: [Proposal] add mapSeparated method on `Iterable`
### Current implementation using an extension method
```dart
extension MapSeparated<S> on Iterable<S> {
  Iterable<T> mapSeparated<T>(T separator, [T Function(S value)? generator]) sync* {
    final iterator = this.iterator;
    if (!iterator.moveNext()) return;
    yield generator?.call(iterator.current) ?? (iterator.current as T);
    while (iterator.moveNext()) {
      yield separator;
      yield generator?.call(iterator.current) ?? (iterator.current as T);
    }
  }
}
```

I chose to add this method as an extension of the Iterable type so I can use it for both List and Itrables. I mostly use this with List type and because of that I have to use either the spread operator to add output in another list or use it with `toList()`. We should discuss if this should stay as it is or we should add this only for `List`.

### Use case:

Sometimes we want that list out our list or iterable with something in between. For example, we have a list of widgets to use in the column with `SizedBox(height: 10)` but we are not using `ListView.separated` for some reason. There this will be useful

Also, it can be useful when we are not sure whether each element of the column is displayed or not based on some condition and we want to add some space between two children.

eg.
```dart
...[
  if (review.content.title.isNotEmpty) ...[
    Text(
      review.content.title!.capitalize,
      style: const TextStyle(fontSize: 12, fontWeight: FontWeight.bold),
    ),
  ],
  if (review.content.review.isNotEmpty) ...[
    Text(
      review.content.review!.capitalize,
      maxLines: 5,
      overflow: TextOverflow.ellipsis,
      style: const TextStyle(fontSize: 12),
    ),
  ],
].mapSeparated<Widget>(const SizedBox(height: 4)),
```

---

Notice that I have used `Iterable<S>`, this can be useful when we have a list of type A and we want type B as output with separators. In the below example `S` is assigned as `({String text, Information info})` and We are using `mapSeparated` method to show all clickable text in `RichText` with `,` in between.

```dart
[
  (text: 'Terms of Use', info: Information.terms),
  (text: 'Privacy', info: Information.privacy),
  (text: 'Delivery', info: Information.delivery),
  (text: 'Auto-Ship', info: Information.autoship),
].mapSeparated(const TextSpan(text: ', '), (value) {
  return TextSpan(
    text: value.text,
    recognizer: value.info.tapRecognizer,
    style: const TextStyle(
      color: AppColor.primary,
      fontWeight: FontWeight.bold,
    ),
  );
}),
```

If anyone has any different views or take on this please share! I was using this with almost all of my projects hence I thought it should be added in here.

### LINKED ISSUE #674: [Proposal] add mapSeparated method on `Iterable`
### Current implementation using an extension method
```dart
extension MapSeparated<S> on Iterable<S> {
  Iterable<T> mapSeparated<T>(T separator, [T Function(S value)? generator]) sync* {
    final iterator = this.iterator;
    if (!iterator.moveNext()) return;
    yield generator?.call(iterator.current) ?? (iterator.current as T);
    while (iterator.moveNext()) {
      yield separator;
      yield generator?.call(iterator.current) ?? (iterator.current as T);
    }
  }
}
```

I chose to add this method as an `extension` of the `Iterable` type so I can use it for both `List` and `Itrables`. I mostly use this with `List` type and because of that I have to use either the spread operator to add output in another `List` or use it with `toList()`. We should discuss if this should stay as it is or we should add this only for `List`.

### Use case:

Sometimes we want that list out our `List` or `Iterable` with something in between. For example, we have a `List` of `Widget` to use in the `Column` or `Row` with `SizedBox(height: 10)` but we are not using `ListView.separated` for some reason. There this can be useful!

Also, it can be useful when we are not sure whether each element of the `Column` or `Row` is displayed or not based on some condition and we want to add some space between two children.

eg.
```dart
...[
  if (review.content.title.isNotEmpty) ...[
    Text(
      review.content.title!.capitalize,
      style: const TextStyle(fontSize: 12, fontWeight: FontWeight.bold),
    ),
  ],
  if (review.content.review.isNotEmpty) ...[
    Text(
      review.content.review!.capitalize,
      maxLines: 5,
      overflow: TextOverflow.ellipsis,
      style: const TextStyle(fontSize: 12),
    ),
  ],
].mapSeparated<Widget>(const SizedBox(height: 4)),
```

---

Notice that I have used `Iterable<S>`, this can be useful when we have a `List` of type `A` and we want `List` of type `B` as output with separators. In the below example `S` is `({String text, Information info})` and We are using `mapSeparated` method to show all clickable `TextSpan` in `RichText` with `,` in between.

```dart
[
  (text: 'Terms of Use', info: Information.terms),
  (text: 'Privacy', info: Information.privacy),
  (text: 'Delivery', info: Information.delivery),
  (text: 'Auto-Ship', info: Information.autoship),
].mapSeparated(const TextSpan(text: ', '), (value) {
  return TextSpan(
    text: value.text,
    recognizer: value.info.tapRecognizer,
    style: const TextStyle(
      color: AppColor.primary,
      fontWeight: FontWeight.bold,
    ),
  );
}),
```

If anyone has any different views or take on this please share! I was using this with almost all of my projects hence I thought it should be added in here.

kishan-dhankecha:
Fixes dart-lang/core#674.

---

### DIFF
--- a/lib/src/iterable_extensions.dart
+++ b/lib/src/iterable_extensions.dart
@@ -56,6 +56,27 @@ extension IterableExtension<T> on Iterable<T> {
     return chosen;
   }
 
+  /// Separates the elements of the iterable with the specified separator.
+  ///
+  /// The [separated] method takes a separator as input and returns a new
+  /// iterable where the elements are separated by the specified separator.
+  ///
+  /// Example:
+  /// ```dart
+  /// final list = [1, 2, 3, 4];
+  /// final separatedList = list.separated(0);
+  /// print(separatedList); // Output: [1, 0, 2, 0, 3, 0, 4]
+  /// ```
+  Iterable<T> separated(T separator) sync* {
+    final iterator = this.iterator;
+    if (!iterator.moveNext()) return;
+    yield iterator.current;
+    while (iterator.moveNext()) {
+      yield separator;
+      yield iterator.current;
+    }
+  }
+
   /// The elements that do not satisfy [test].
   Iterable<T> whereNot(bool Function(T element) test) =>
       where((element) => !test(element));
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## 1.19.0-wip
 
 - Adds `shuffled` to `IterableExtension`.
+- Adds `separated` to `IterableExtension`.
 - Shuffle `IterableExtension.sample` results.
 - Fix `mergeSort` when the runtime iterable generic is a subtype of the static
   generic.
--- a/test/extensions_test.dart
+++ b/test/extensions_test.dart
@@ -850,6 +850,25 @@ void main() {
               ]);
         });
       });
+      group('.separated', () {
+        test('empty', () {
+          expect(iterable([]).separated(unreachable), isEmpty);
+        });
+
+        test('single', () {
+          expect(iterable([1]).separated(0), [1]);
+        });
+
+        test('multiple', () {
+          expect(iterable([1, 2, 3]).separated(0), [1, 0, 2, 0, 3]);
+          expect(iterable(['a', 'b']).separated('-'), ['a', '-', 'b']);
+        });
+
+        test('dynamic', () {
+          expect(iterable<Object>([1, '2', 3]).separated(0), [1, 0, '2', 0, 3]);
+          expect(iterable<Object>(['a', 'b']).separated(0), ['a', 0, 'b']);
+        });
+      });
       group('none', () {
         test('empty', () {
           expect(iterable([]).none(unreachable), true);


kevmoo:
I'd leave the review to @devoncarew and/or @lrhn

kevmoo (approved):
LGTM – just want one more nod on this. We should be mindful when adding things we'll need to support "forever".

lib/src/iterable_extensions.dart
```diff
@@ -56,6 +56,27 @@ extension IterableExtension<T> on Iterable<T> {
...
+  /// ```
+  Iterable<T> separated(T separator) sync* {
+    final iterator = this.iterator;
+    if (!iterator.moveNext()) return;
+    yield iterator.current;
+    while (iterator.moveNext()) {
+      yield separator;
+      yield iterator.current;
+    }
+  }
```
lrhn, line 78: I don't think this version is flexible enough.
The use-cases I've heard mentioned won't all be satisfies by inserting the *same* value every time, and the typing might not work out.

I'd *at least* want a version where the separator is generated each time.
Then it could be generated based on position, or on adjacent values.

Something like these would cover those:
```dart
Iterable<T> separatedByValue(T separator) { ... }
Iterable<T> separatedBy(T Function() separator) { ... }
```
and possibly also:
```dart
// Position is index of next element, from 1 to `lenght` - 1.
Iterable<T> separatedByIndex(T Function(int position) separator) { ... }
Iterable<T> separatedByElements(T Function(T before, T after) separator) { ... }
```
The `separatedBy` and `separatedByIndex` might be combined into one, it's easy to ignore an argument, and cheap to increment an integer.

I'd expect the separator to not always have the same type as the elements being separated. (A `List<TextWidgets>` cannot be separated by `SeparatorWidget`s, one would have to first copy the text widgets into a `List<Widget>` and then separate them)

That makes a top-level function more useful (or stastic function or constructor, if we could add such using extensions):
```dart
  static Iterable<S> separatedBy<S>(Iterable<S> elements, S Function(int position)  separator) { ... }
  static Iterable<S> separatedByValue<S>(Iterable<S> elements, S  separator) { ... }
  static Iterable<S> separatedByElements<S, T extends S>(Iterable<T> elements, S Function(T before, T after)  separator) { ... }
```
Maybe we can make the separator optional, like: 
```dart
  static Iterable<S> separatedOptionalBy<S>(Iterable<S> elements, S? Function(int position)  separator) { ... }
  static Iterable<S> separatedOptionalByElements<S, T extends S>(Iterable<T> elements, S? Function(T before, T after)  separator) { ... }
```
where returning `null` means no separator.
(Consider: `separatedOptionalByElements(pathSegments, (before, after) => before.endsWith('/') || after.startsWith('/') ? null : '/')` where we only insert a separator where it's needed. Maybe this is too specialized, and a `join` with  conditional separator would be more relevant.)

mosuem:

PR #347, opened 2024-06-07
`CanonicalizedMap`: added constructor `fromEntries`.

gmpassos:
Currently, the only way to create a `CanonicalizedMap` is from a fully constructed `Map`.

With `fromEntries`, a common use case, we can avoid the creation of an intermediate `Map`.

### Real-world case:

Avoid unnecessary overhead as seen here:

https://github.com/dart-lang/shelf/blob/master/pkgs/shelf/lib/src/headers.dart#L32

https://github.com/dart-lang/http_parser/blob/master/lib/src/case_insensitive_map.dart#L13

### Reason:

Improve Dart benchmark performance:

https://sharkbench.dev/web

---

- [✅ ] I’ve reviewed the contributor guide and applied the relevant portions to this PR.

### DIFF
--- a/test/canonicalized_map_test.dart
+++ b/test/canonicalized_map_test.dart
@@ -205,7 +205,7 @@ void main() {
     });
   });
 
-  group('CanonicalizedMap.from', () {
+  group('CanonicalizedMap.fromEntries', () {
     test('canonicalizes its keys', () {
       var map = CanonicalizedMap.fromEntries(
           {'1': 'value 1', '2': 'value 2', '3': 'value 3'}.entries, int.parse);
--- a/test/canonicalized_map_test.dart
+++ b/test/canonicalized_map_test.dart
@@ -205,6 +205,24 @@ void main() {
     });
   });
 
+  group('CanonicalizedMap.from', () {
+    test('canonicalizes its keys', () {
+      var map = CanonicalizedMap.fromEntries(
+          {'1': 'value 1', '2': 'value 2', '3': 'value 3'}.entries, int.parse);
+      expect(map['01'], equals('value 1'));
+      expect(map['02'], equals('value 2'));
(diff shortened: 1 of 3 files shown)


gmpassos:
FYI: @kevmoo

kevmoo (approved):
Love it!  Any notes @devoncarew @natebosch ?

test/canonicalized_map_test.dart
```diff
@@ -205,6 +205,24 @@ void main() {
     });
   });
 
+  group('CanonicalizedMap.from', () {
```
jtmcdole, line 208: .fromEntries

kevmoo:
@gmpassos – LOVE THIS work. We're looking at a few other PRs and should have this published on Monday.

I'll let you follow-up with the `shelf` PR. Any idea about how improved the perf is here?

gmpassos:
> @gmpassos – LOVE THIS work. We're looking at a few other PRs and should have this published on Monday.
> 
> 
> 
> I'll let you follow-up with the `shelf` PR. Any idea about how improved the perf is here?

Once this reaches pub.dev I will push a PR on 'http_parser', then one on 'shelf'.

It's all about the tests and merge "ping pong" delay.

@kevmoo,
Thanks for the fast response.

Best regards.

gmpassos:
About the performance improvement: I'm conducting many benchmarks using our own API (real-world) code. One bottleneck is the handling of case-insensitive headers. We will continue to check for other "hotspots" once the new fixes are in production.

kevmoo:
https://pub.dev/packages/collection/versions/1.19.0 – this is published @gmpassos !

What's next? Here to help!

gmpassos:
> What's next? Here to help!

`http_parser` PR:

https://github.com/dart-lang/http_parser/pull/99

PR #349, opened 2024-06-10
Require Dart 3.4, test wasm on 3.4, prepare release

kevmoo:


kevmoo:
We'll want to wait to publish until after we've synced with the SDK and g3

devoncarew (approved):
This has landed in the sdk repo and google3.

PR #357, opened 2024-08-02
Fix issue 355 added get frequency getter to iterable 

pvlKryu:
[yes] I’ve reviewed the contributor guide and applied the relevant portions to this PR.

Description:
This Pull Request introduces a new extension getter frequencies for the Iterable class. The countFrequency method returns a map where the keys are the unique elements of the iterable and the values are the counts of those elements. This addition provides an easy and efficient way to count the frequency of elements in any iterable.

Changes:
- Added frequencies getter extension to lib/src/iterable_extensions.dart
- Implemented comprehensive tests for the frequencies getter in test/extensions_test.dart. The tests cover various subclasses of Iterable including List, Set, and Queue.

Relevant Issues:
This PR addresses [Issue dart-lang/core#679](https://github.com/dart-lang/core/issues/679).

Thank you!

### DIFF
--- a/lib/src/iterable_extensions.dart
+++ b/lib/src/iterable_extensions.dart
@@ -602,11 +602,16 @@ extension IterableExtension<T> on Iterable<T> {
     }
   }
 
-  /// Returns a map where the keys are the unique elements of the iterable
-  /// and the values are the counts of those elements.
+  /// The count of occurrences of each element.
   ///
-  /// for example, ['a', 'b', 'b', 'c', 'c', 'c'].frequencies;
-  /// returns {'a': 1, 'b': 2, 'c': 3}.
+  /// The map has an entry for each distinct element of this iterable,
+  /// as determined by `==`, where the value is the number of eelements
+  /// in this iterable which are equal to the key. 
+  /// If there are elements that are equal, but not identical, its unspecified
+  /// which of the elements is used as the key.
+  ///
+  /// For example `['a', 'b', 'c', 'b', 'c', 'c'].frequencies` 
+  /// is a map with entries like `{'a': 1, 'b': 2, 'c': 3}`.
   Map<T, int> get frequencies {
     final frequencyMap = <T, int>{};
     for (var item in this) {
--- a/lib/src/iterable_extensions.dart
+++ b/lib/src/iterable_extensions.dart
@@ -597,12 +597,15 @@ extension IterableExtension<T> on Iterable<T> {
   /// for example, ['a', 'b', 'b', 'c', 'c', 'c'].frequencies;
   /// returns {'a': 1, 'b': 2, 'c': 3}.
   Map<T, int> get frequencies {
-    var frequencyMap = <T, int>{};
+    final frequencyMap = <T, int>{};
     for (var item in this) {
-      frequencyMap[item] = (frequencyMap[item] ?? 0) + 1;
+      frequencyMap.update(item, _increment, ifAbsent: _one);
     }
     return frequencyMap;
   }
+
+  static int _increment(int value) => value + 1;
+  static int _one() => 1;
 }
 
 /// Extensions that apply to iterables with a nullable element type.
--- a/lib/src/iterable_extensions.dart
+++ b/lib/src/iterable_extensions.dart
@@ -601,6 +601,19 @@ extension IterableExtension<T> on Iterable<T> {
       yield slice;
     }
   }
+
+  /// Returns a map where the keys are the unique elements of the iterable
+  /// and the values are the counts of those elements.
+  ///
+  /// for example, ['a', 'b', 'b', 'c', 'c', 'c'].countFrequency()
+  /// returns {'a': 1, 'b': 2, 'c': 3}. will it works?
+  Map<T, int> countFrequency() {
+    var frequencyMap = <T, int>{};
+    for (var item in this) {
+      frequencyMap[item] = (frequencyMap[item] ?? 0) + 1;
+    }
+    return frequencyMap;
+  }
 }
 
 /// Extensions that apply to iterables with a nullable element type.
--- a/lib/src/iterable_extensions.dart
+++ b/lib/src/iterable_extensions.dart
@@ -604,14 +604,28 @@ extension IterableExtension<T> on Iterable<T> {
 
   /// The count of occurrences of each element.
   ///
-  /// The map has an entry for each distinct element of this iterable,
-  /// as determined by `==`, where the value is the number of eelements
-  /// in this iterable which are equal to the key. 
-  /// If there are elements that are equal, but not identical, its unspecified
-  /// which of the elements is used as the key.
+  /// The map contains an entry for each unique element of this iterable,
+  /// as determined by `==`.
+  /// The value for each key is the number of times that element 
+  /// appears in the iterable.
   ///
-  /// For example `['a', 'b', 'c', 'b', 'c', 'c'].frequencies` 
-  /// is a map with entries like `{'a': 1, 'b': 2, 'c': 3}`.
+  /// If there are elements that are equal (`==`), 
+  /// but not identical (`identical`), 
+  /// it is unspecified which of the elements is used as the key in the map. 
+  /// For example, if there are multiple lists with the same content, 
+  /// the map will only keep one of them as a key.
+  ///
+  /// Example:
+  /// ```dart
+  /// ['a', 'b', 'c', 'b', 'c', 'c'].frequencies; 
+  /// Returns: {'a': 1, 'b': 2, 'c': 3}.
+  /// ```
+  ///
+  /// Note: This method uses `==` to compare elements. 
+  /// For collections # `List`, `Set`, or `Map`, deep equality is not checked. 
+  /// If you need deep equality (e.g., nested lists),
+  /// consider using a custom equality mechanism.
+
   Map<T, int> get frequencies {
     final frequencyMap = <T, int>{};
     for (var item in this) {
--- a/lib/src/iterable_extensions.dart
+++ b/lib/src/iterable_extensions.dart
@@ -605,8 +605,8 @@ extension IterableExtension<T> on Iterable<T> {
   /// Returns a map where the keys are the unique elements of the iterable
   /// and the values are the counts of those elements.
   ///
-  /// for example, ['a', 'b', 'b', 'c', 'c', 'c'].countFrequency()
-  /// returns {'a': 1, 'b': 2, 'c': 3}.
+  /// For example, `['a', 'b', 'b', 'c', 'c', 'c'].countFrequency()`
+  /// returns `{'a': 1, 'b': 2, 'c': 3}`.
   Map<T, int> countFrequency() {
     var frequencyMap = <T, int>{};
     for (var item in this) {
--- a/lib/src/iterable_extensions.dart
+++ b/lib/src/iterable_extensions.dart
@@ -57,7 +57,8 @@ extension IterableExtension<T> on Iterable<T> {
   }
 
   /// The elements that do not satisfy [test].
-  Iterable<T> whereNot(bool Function(T element) test) => where((element) => !test(element));
+  Iterable<T> whereNot(bool Function(T element) test) =>
+      where((element) => !test(element));
 
   /// Creates a sorted list of the elements of the iterable.
   ///
@@ -81,7 +82,8 @@ extension IterableExtension<T> on Iterable<T> {
   ///
   /// The elements are ordered by the [compare] [Comparator] of the
   /// property [keyOf] of the element.
-  List<T> sortedByCompare<K>(K Function(T element) keyOf, Comparator<K> compare) {
+  List<T> sortedByCompare<K>(
+      K Function(T element) keyOf, Comparator<K> compare) {
     var elements = [...this];
     mergeSortBy<T, K>(elements, keyOf, compare);
     return elements;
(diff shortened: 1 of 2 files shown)


pvlKryu:
@mateusfccp Hey Mateus, can you approve?

pvlKryu:
@kevmoo can you please review this one?

mateusfccp:
> @mateusfccp Hey Mateus, can you approve?

I can't, I'm not a member of the team.

lib/src/iterable_extensions.dart
```diff
@@ -601,6 +601,19 @@ extension IterableExtension<T> on Iterable<T> {
       yield slice;
     }
   }
+
+  /// Returns a map where the keys are the unique elements of the iterable
+  /// and the values are the counts of those elements.
+  ///
+  /// For example, `['a', 'b', 'b', 'c', 'c', 'c'].countFrequency()`
+  /// returns `{'a': 1, 'b': 2, 'c': 3}`.
+  Map<T, int> countFrequency() {
```
lrhn, line 610: Would probably make it `get frequencies` instead of a function.

lib/src/iterable_extensions.dart
```diff
@@ -601,6 +601,19 @@ extension IterableExtension<T> on Iterable<T> {
...
+    var frequencyMap = <T, int>{};
+    for (var item in this) {
+      frequencyMap[item] = (frequencyMap[item] ?? 0) + 1;
```
lrhn, line 613: Can use `update`, which avoids doing two lookups per item.
```dart
   frequencyMap.update(item, _increment, ifAbsent: _one);
```
with top-level/static helpers:
```dart
  int _increment(int value) => value + 1;
  int _one() => 1;
```

lib/src/iterable_extensions.dart
pvlKryu: thank you, I agree this way is better, please check my changes

lib/src/iterable_extensions.dart
pvlKryu: thank you, I agree this way is better, please check my changes

lib/src/iterable_extensions.dart
```diff
@@ -601,6 +601,22 @@ extension IterableExtension<T> on Iterable<T> {
       yield slice;
     }
   }
+
+  /// Returns a map where the keys are the unique elements of the iterable
```
lrhn, line 605: Single line first paragraph. Format as sentences.
Don't use `Returns` for a getter. (Don't start any DartDoc with it, there is always a better word).

I'd say:
```dart
  /// The count of occurrences of each element.
  ///
  /// The map has an entry for each distinct element of this iterable,
  /// as determined by `==`, where the value is the number of eelements
  /// in this iterable which are equal to the key. 
  /// If there are elements that are equal, but not identical, its unspecified
  /// which of the elements is used as the key.
  ///
  /// For example `['a', 'b', 'c', 'b', 'c', 'c'].frequencies` 
  /// is a map with entries like `{'a': , 'b': 2, 'c': 3}`.
```

test/extensions_test.dart
```diff
@@ -1352,6 +1353,84 @@ void main() {
...
+        expect(frequencyMap, {});
+      });
+
+      test('should handle single element Set', () {
+        var set = {42};
+        var frequencyMap = set.frequencies;
+        expect(frequencyMap, {42: 1});
+      });
+
+      test('should return correct frequency map for Queue of integers', () {
```
lrhn, line 1407: It's probably overkill to go throguh all kinds of iterables.
I'd rather have more diversity in the elements.

For example having elements that are equal, but not identical, or having elements that are records.
It's the building of the map that is tricky, iterating the iterable will probably work.

lrhn:
@natebosch WDYT. Is this worth adding?
(it's pretty simple, so no big complexity add, and it's something that is often hard for people to write efficiently.)

natebosch:
> @natebosch WDYT. Is this worth adding?

Yes, I think it is worth adding. It's not as general as `groupFoldBy` without the `keyOf` callback, but you can accomplish the same with a `.map` call. Getting frequencies by `==` is common enough to have the specialized implementation.

lib/src/iterable_extensions.dart
pvlKryu: corrected

test/extensions_test.dart
pvlKryu: added `get frequencies tests extended `group

lib/src/iterable_extensions.dart
```diff
@@ -601,6 +601,41 @@ extension IterableExtension<T> on Iterable<T> {
...
+  /// the map will only keep one of them as a key.
+  ///
+  /// Example:
+  /// ```dart
+  /// ['a', 'b', 'c', 'b', 'c', 'c'].frequencies; 
+  /// Returns: {'a': 1, 'b': 2, 'c': 3}.
+  /// ```
+  ///
+  /// Note: This method uses `==` to compare elements. 
+  /// For collections # `List`, `Set`, or `Map`, deep equality is not checked.
```
lrhn, line 625: Not even shallow equality is checked, two lists are equal only if they are the same instance.

lib/src/iterable_extensions.dart
```diff
@@ -601,6 +601,41 @@ extension IterableExtension<T> on Iterable<T> {
...
+  /// The map contains an entry for each unique element of this iterable,
+  /// as determined by `==`.
+  /// The value for each key is the number of times that element 
+  /// appears in the iterable.
+  ///
+  /// If there are elements that are equal (`==`), 
+  /// but not identical (`identical`), 
+  /// it is unspecified which of the elements is used as the key in the map. 
+  /// For example, if there are multiple lists with the same content, 
+  /// the map will only keep one of them as a key.
```
lrhn, line 616: Not a good example, though, since list equality is identity based.

test/extensions_test.dart
```diff
@@ -1352,6 +1353,122 @@ void main() {
...
+        var queue = Queue<int>();
+        queue.add(42);
+        var frequencyMap = queue.frequencies;
+        expect(frequencyMap, {42: 1});
+      });
+    });
+
+    group('get frequencies tests extended', () {
+      test('list of equal but not identical strings', () {
+        var list = ['apple', String.fromCharCodes('apple'.codeUnits)];
```
lrhn, line 1437: String is not a good choice for this test because those will be identical when compiled for web.

Consider adding a class:
```dart
class Wrap {
  final int id;
  final int value;
  Wrap(this.id, this.value);
  int get hashCode => value.hashCode;
  bool operator ==(Object other) => other is Wrap && value == other.value;
  String toString() => '(#$id:$value)';
}
```
and using that to create distinct-but-equal values:
```dart
  var list = [Wrap(0, 1), Wrap(1, 1), Wrap(0, 2)];
  var frequenceyMap = list.frequencies;
  expect(frequencyMap[list[0]], 2);
  expect(frequencyMap[list[1]], 2);
  expect(frequencyMap[list[2]], 1);
```
or similar.

test/extensions_test.dart
```diff
@@ -1352,6 +1353,122 @@ void main() {
...
+        expect(frequencyMap, {(1, 'a'): 3, (2, 'b'): 1});
+      });
+
+      test('list with elements that are objects', () {
+        var list = [const MyObject(1), const MyObject(1), const MyObject(2)];
+        var frequencyMap = list.frequencies;
+        expect(frequencyMap, {const MyObject(1): 2, const MyObject(2): 1});
+      });
+
+      test('list with equal numbers but different types', () {
```
lrhn, line 1454: (Good test. There is no way around web considering them all identical here.)

test/extensions_test.dart
```diff
@@ -1352,6 +1353,122 @@ void main() {
...
+      });
+
+      test('list of records', () {
+        var list = [(1, 'a'), (1, 'a'), (2, 'b'), (1, 'a')];
+        var frequencyMap = list.frequencies;
+        expect(frequencyMap, {(1, 'a'): 3, (2, 'b'): 1});
+      });
+
+      test('list with elements that are objects', () {
+        var list = [const MyObject(1), const MyObject(1), const MyObject(2)];
```
lrhn, line 1449: Don't use `const` here, because that makes the objects actually identical, so it doesn't check equality being different from identitiy.
Or have one `const` and one not.

test/extensions_test.dart
```diff
@@ -1352,6 +1353,122 @@ void main() {
...
+      test('should return correct frequency map for List of strings', () {
+        var list = ['a', 'b', 'b', 'c', 'c', 'c'];
+        var frequencyMap = list.frequencies;
+        expect(frequencyMap, {'a': 1, 'b': 2, 'c': 3});
+      });
+
+      test('should handle empty List', () {
+        var list = [];
+        var frequencyMap = list.frequencies;
+        expect(frequencyMap, {});
```
lrhn, line 1372: Consider `isEmpty` as matcher. I *think* it works for maps too, and it should give an error message like "is not empty".

test/extensions_test.dart
```diff
@@ -1352,6 +1353,122 @@ void main() {
         expect(l3.toList(), [4, 5]);
       });
     });
+    group('get frequencies tests', () {
```
lrhn, line 1356: Just `'frequenices'`, it's cleaner.

The names of groups and tests are concatenated with a space between them, so they can rely on context. Things like "should be correct" is a given for tests, so it doesn't add value. Consider:
```dart
    group('frequencies', () { 
      test('of integers', () { ... });
      test('of strings', () { ... });
      test('of empty iterable', () { ... });
      test('of empty single element iterable', () { ... });
      group('where equality is not identity,', () {
        test('of records', () { ... });
        /// ...
      });
   });
```
These names are only important when a test fails.
At that point, they should give enough context to make the error message meaningful.

Take "frequencies of single element iterable" with an error (if the result was empty) of 
```
  Expected: {42: 1}
    Actual: {}
     Which: has different length and is missing map key <42>
```
With that headline, it's fairly clear where and what the problem is.

test/extensions_test.dart
```diff
@@ -1352,6 +1353,122 @@ void main() {
...
+      });
+
+      test('should return correct frequency map for Set of strings', () {
+        // ignore: equal_elements_in_set
+        var set = {'a', 'b', 'b', 'c', 'c', 'c'};
+        var frequencyMap = set.frequencies;
+        expect(frequencyMap, {'a': 1, 'b': 1, 'c': 1});
+      });
+
+      test('should handle empty Set', () {
```
lrhn, line 1395: I would *drop* the tests for sets and queues unless there is a reason to believe that the implementation treats them differently (which it doesn't).

We're not testing iteration, that is assumed to work, so all that matters is the sequence of values we get by iterating, and a `List` is fine for that.

(If we start trying to detect lists or sets, to do more efficient iteration or maybe assuming that sets cannot have duplicates so we don't need to count - which isn't true because sets can have non-standard equality - then there is reason to test with, and without, those types specifically.)

test/extensions_test.dart
```diff
@@ -1352,6 +1353,122 @@ void main() {
...
+    });
+
+    group('get frequencies tests extended', () {
+      test('list of equal but not identical strings', () {
+        var list = ['apple', String.fromCharCodes('apple'.codeUnits)];
+        var frequencyMap = list.frequencies;
+        expect(frequencyMap, {'apple': 2});
+      });
+
+      test('list of records', () {
```
lrhn, line 1442: (So for naming, I'd drop the "list of" here, what is important is that it's records being compared, and it being a list isn't important to the test.
Maybe just `test('of records', () { ... });` with the changes above.)

lrhn (commented):
Looking good, but if doing shallow comparison of lists is a goal, we need to parameterize with an equality.

pvlKryu:
@lrhn Thank you very much for the quick feedback and for the detailed and helpful comments. I need some time to review everything you mentioned and correct it. I will let you know soon!

lib/src/iterable_extensions.dart
mateusfccp: How expensive is a lookup? Shouldn't it `O(1)`? So we are basically avoiding `O(2)`?

lrhn: Yes, we are avoiding O(2) and lowering the constant factor. I hope, that assumes the map is actually implemented efficiently, but it's at least *possible*.

Also, for degenerate cases, hash table lookup is linear. And quick-sort is quadratic. We always hope to not hit those cases, but when we do, it's better to have a low constant.

For library code that other people will use, especially low-level code like this, efficiency is important.
If all your helper libraries are 10% slower than necessary, then your entire application is *at least* 10% slower than necessary. (If helper libraries use helper libraries, it may compound.)
And there is nothing you can do about it.

So when you provide code for others to depend on, to build their algorithms and applications on, you should *never* say "a constant factor doesn't matter", because to end users, the difference between 15ms and 17ms may be a missed animation frame, and that *matters*.

(So, that's my pocket philosophy about writing code for others: The only person who can say whether performance is "good enough" is the application writer. A helper library doesn't know that. It may choose *how* it optimizes, for speed or size or a tradeoff, but it shouldn't say "it's good enough" and leave easy optimizations on the table. Every level below the application has to do their best and not be part of the problem.)

The one thing we should actually do is to *measure*. If `operator[]` and `operator[]=` are so heavily optimizied that they beat `update` even if they do the same lookup twice, then we need to optimize `update` too. I want to choose the operation that has the best *possible* performance, rather than using workarounds that are faster right now, and never come back and check if they stay that way. If `update` isn't faster today, we should file an issue, so that it is tomorrow.

(I can see that I flip between pragmatic and ideological reasoning - be fast, because it matters, but don't be fast by doing it the wrong way, even if it *is* faster today. Tradeoffs. It's tradeoffs all the way down. So reasonable people can disagree on where to trade what. )

mateusfccp: I'm not disagreeing, just wanted to understand the reasoning.

Thanks for the explanation.

lib/src/iterable_extensions.dart
natebosch: I think this is "shallow" as it relates to `this`, not as it relates to the elements of `this` that are also collection typed.

mosuem:

PR #360, opened 2024-09-23
Implement `firstWhereType` and `firstWhereTypeOrNull`

### LINKED ISSUE #352: [Proposal] Add `firstWhereType`
We already have `whereType`, but it would be great to get the first occurrence of the given type.

Here's a use case where this could come in hand.

```dart
final results = await Future.wait(
  [
    _metadataService.getTableMetadata(tableId),
    _dataService.getTableData(tableId),
  ],
  eagerError: true,
);

final metadata = results.whereType<TableMetadata>().first;
final data = results.whereType<TableRawData>().first;
```

It could be replaced with

```dart
final metadata = results.firstWhereType<TableMetadata>();
final data = results.firstWhereType<TableRawData>();
```

mateusfccp:
Adds `firstWhereType` and `firstWhereTypeOrNull`.
Fixes #352.

---

<details>

- Most changes should add an entry to the changelog and may need to [rev the pubspec package version](https://github.com/dart-lang/sdk/blob/main/docs/External-Package-Maintenance.md#making-a-change).

</details>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 1.20.0-wip
+
+- Adds `firstWhereType` to `IterableExtension`.
+- Adds `firstWhereTypeOrNull` to `IterableExtension`.
+
 ## 1.19.0
 
 - Adds `shuffled` to `IterableExtension`.
--- a/lib/src/iterable_extensions.dart
+++ b/lib/src/iterable_extensions.dart
@@ -250,6 +250,28 @@ extension IterableExtension<T> on Iterable<T> {
     return result;
   }
 
+  /// The first element with type [T].
+  ///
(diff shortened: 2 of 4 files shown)


lrhn:
Not sure these combinations are worth having as specialized functions. Combining the `whereType` and `first`/`firstWhere` primitives works well and has no obvious usability pitfalls (other than finding them, if starting from `.firstWhere((x) => x is T)`).

If you are using an iterable to begin with (and not, say, a `List`), you're probably not trying to get absolutely optimal performance. I wouldn't worry about the overhead of one extra iterable.

@natebosch WDYT?

natebosch:
I agree. These are too specialized and `whereType<Foo>().firstOrNull` is a good pattern to use. I'd recommend defining these locally if you prefer this style.

